PR #60, opened 2021-04-21
Plugin right click tray

SlimRunner:
The syntax of the plugin is a bit messy because my inexperience with typescript, but it works exactly as it should when disabled and enabled.

You might want to review the order of the toggles. I put it at the bottom.

In short what the plugin does is enable the user to right click the expression's settings bubble instead of having to long hold.


jared-hughes (changes requested):
Please merge https://github.com/SlimRunner/right-click-tray/pull/1. Otherwise, it looks great

jared-hughes (approved):
Looks like the updated submodule works. I can't wait till #61 and hopefully a better plugin system.

PR #94, opened 2021-05-17
🔨 Refactor wolfram2desmos to align with Desmodder guidelines

SlimRunner:


jared-hughes:
Looks good! Loving that it can be enabled/disabled at will.

PR #138, opened 2021-05-31
🐛 [wolfram2desmos] Bugfix for erroneous csv parsing

SlimRunner:
I fixed the bug addressed [on issue 17](https://github.com/Heavenira/wolfram2desmos/issues/17) in the wolfram2desmos repository.

I made the following changes:

- Added TableParse as a new util for plugins.
- Used TableParse to prevent csv data being parsed as ascii-math
- The wolfram2desmos function now returns null when input is not ascii-math
- Added new restriction in wolfram2desmos to discriminate single row csv data (which TableParse fails to capture)

I added the TableParse as a util following guideline on this comment: https://github.com/DesModder/DesModder/blob/570c9e160aabbeced815561115b729c7098e680c/src/desmodder.ts#L23-L25
If you need changes in there let me know.

### DIFF
--- a/src/desmodder.ts
+++ b/src/desmodder.ts
@@ -39,6 +39,7 @@ export {
 } from "DCGView";
 export {
   EvaluateSingleExpression,
+	TableParse,
   pollForValue,
   mergeClass,
   MaybeClassDict,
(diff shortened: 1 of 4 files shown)


SlimRunner:
I did not test thoroughly and actually introduced a syntax error and a bug with this patch I'm about to amend a fix to this.

SlimRunner:
Due a recent refactor from @Heavenira on wolfram2desmos discussed in #142 and the lack of discussion here I will close this pull request.

Issue #131, opened 2021-05-25
Large simulation update rules get cut off in the preview


jared-hughes:
See https://www.desmos.com/calculator/b45cwl7vb6 for an example. This has two harmful effects: a scrollbar forms on the bottom, and the math gets cut off on the side.

![image](https://user-images.githubusercontent.com/20214911/119442842-6a2c1400-bcdd-11eb-84f9-5595363827bd.png)

Preferred fix: fade the equation into nothingness like the normal simulation object:

![image](https://user-images.githubusercontent.com/20214911/119443078-ce4ed800-bcdd-11eb-9534-6aaefb8a40e3.png)

PR #139, opened 2021-05-31
🐛 [video-creator] Fix overflow simulation rule preview

SlimRunner:
This is a fix to issue #131 which causes the preview of the simulation rules to clip off the popover.

I mirrored the some of the same CSS rules that are already used in Desmos' simulation expressions. One thing that I changed is the max length of the assignment field and the rule field because Desmos uses 120 pixels for both but this still causes overflow in Desmodder. I opted for applying a first and last child rule to them with 70 and 120 pixels respectively.

If you want changes briefly tell me what you want differently, and I will look into it.

### DIFF
--- a/src/plugins/video-creator/components/SimulationPicker.css
+++ b/src/plugins/video-creator/components/SimulationPicker.css
@@ -9,3 +9,30 @@
   padding: 0 8px;
   transform: translateY(1px);
 }
+
+.video-creator-simulation-rules
+  .dcg-clickable-property-row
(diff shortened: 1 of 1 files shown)


SlimRunner:
Found an issue myself right now. I need to disable the fade when there is no overflow detected.

SlimRunner:
I cannot find a way to make the overflow detection happen. I might revisit this later, but in the mean time I will close this pull request since the changes I have made so far are trivial.

PR #182, opened 2021-08-02
Fixed lookahead crash error

Heavenira:
Removed indexing of `input` to prevent possible crash


jared-hughes:
Context: this successfully fixes the crash when `(1)/sqrt(x)` is pasted. Merging

PR #214, opened 2021-10-03
Piecewise statements implemented.

Heavenira:
This should clear up some issues when pasting piecewise expressions. Not tested formally yet.


jared-hughes:
This fails in compilation.

A few errors:

```
Line 106:       TS2454: Variable 'i' is used before being assigned.
Line 107:       TS2454: Variable 'bracket' is used before being assigned.
Line 788:       TS1005: '}' expected.
```

Might just be able to move the declarations of `i` and `bracket` higher in the file to avoid some of those errors. Not sure about the missing closing bracket though, but please make sure the code can compile.

jared-hughes:
Currently converts `{x>0: 1/2, sin(3)}` to `\left\{x>0:\frac{1}{2,}\sin\left(3\right)\right\}`, which is not appropriate behavior.

PR #320, opened 2022-01-20
[] bracket support + function exponents fix'd

Heavenira:
[] brackets like `a[2]` will become fancier +
trig exponents work correctly now.

### DIFF
--- a/src/plugins/wolfram2desmos/wolfram2desmos.ts
+++ b/src/plugins/wolfram2desmos/wolfram2desmos.ts
@@ -49,14 +49,14 @@ export function isIllegalASCIIMath(input: string) {
     //console.warn("Input has " + (count(/\}/g) - count(/\{/g)) + " more '}' characters than '{' characters");
     return false;
   }
-  if (count(/\[/g) > count(/\]/g)) {
+  if (count(/\[/g) != count(/\]/g)) {
     //console.warn("Input has " + (count(/\[/g) - count(/\]/g)) + " more '[' characters than ']' characters");
     return false;
   }
-  if (count(/\[/g) < count(/\]/g)) {
+  /*if (count(/\[/g) < count(/\]/g)) {
     //console.warn("Input has " + (count(/\]/g) - count(/\[/g)) + " more ']' characters than '[' characters");
     return false;
-  }
+  }*/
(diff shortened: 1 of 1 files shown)


Heavenira:
Whoops... I am not sure why that failed. If this is TypeScript related, it is sadly beyond me.

Scoder12:
You could reduce it to one if statement using `!==`.
The error is due to code formatting. Try running `npm run prettier`.

Heavenira:
> You could reduce it to one if statement using `!==`. The error is due to code formatting. Try running `npm run prettier`.

Which line can be reduced? I am not sure where this refers to.

Scoder12:
Instead of checking both < and >, you can just check whether they are not equal, like `count() !== count()`

Heavenira:
Wait, do I have to do that with every single < and > check? This must be a new criteria that @jared-hughes implemented.

Scoder12:
No, logically `a > b OR a < b` is the same as `a != b`

Heavenira:
I'm a little bit miffed that GitHub is flagging this as an error. I understand that `a > b OR a < b` is the same as `a != b`. Although both they `return false;` identically, the `console.warn` still reveals a certain aspect of how it failed.
Should I just remove the comments entirely, and do this merge with `a != b` notation?

Scoder12:
Don't worry, it's not an error, just a nitpick. If you want specific messages then it makes sense. The error is because you need to run prettier.

Heavenira (commented):
Reverted to initial fix, as I don't particularly mind the `pretty` error

jared-hughes:
The issue was that you used tabs instead of spaces, which prettier doesn't like. You'll want to install the Editorconfig and/or Prettier extensions to stay consistent with spaces in the repository.

SlimRunner:
My two cents about setting up Prettier in case this is what's giving you issues, @Heavenira. I found it a lot easier setting up Prettier on VS code than on Atom. I had to install like 5 dependencies to get prettier working on Atom while in VS code I simply installed the prettier extension and done.

Heavenira:
@SlimRunner Just installed! Thanks for directing me to VSCode's extension. I feared that `npm` would be required for me to get it working, (which frankly I'm not down to do :D )

Issue #318, opened 2022-01-15
Suggestion: Info display


GanerCodes:
An option to enable certain info such as mouse location on graph, framerate (I know you can use url tricks but a toggle would be better), etc. would be nice to have

jared-hughes:
Could you clarify what you mean by mouse location on graph? Do you want a variable that gets continuously updated with the position of the mouse?

The framerate + performance is already planned, e.g. https://github.com/DesModder/DesModder/issues/45 and https://github.com/DesModder/DesModder/issues/307.

Do you have any other suggestions collapsed into "etc."?

GanerCodes:
> Could you clarify what you mean by mouse location on graph? Do you want a variable that gets continuously updated with the position of the mouse?

I mean sorta like the fps counter you can see the exact coordinate your mouse location is on the graph, it's faster and more accurate than looking at the gridlines

jared-hughes:
Oh that's a great idea! In some situations, it can be much better than messing around with dragging points.

I'm currently envisioning a HUD where you can choose from various readouts, including:

- xy mouse position
- FPS
- Time in Worker

GanerCodes:
That sounds awesome, thanks!

jared-hughes:
Resolved by #412 (performance view). XY mouse position doesn't help too much because draggable points also work, so that sub-request is declined for now.

PR #412, opened 2022-08-18
Add Performance Info Plugin

### LINKED ISSUE #45: Speed profiling
Would apply to profiling as slider changes & as simulation proceeds. (See https://discord.com/channels/655972529030037504/663607163532476426/828865858956296192; allows checking average milliseconds per tick, but that could be affected by other factors).  It appears that extensions can view browser performance internals (https://chrome.google.com/webstore/detail/performance-analyser/djgfmlohefpomchfabngccpbaflcahjf), which could allow for more accurate profiling, but most accurate would involve accessing the worker directly.

Related: #41

FenyLabs:
Performance info plugin that allows you to... view your graph's performance. Pretty minimal plugin for now because of school, but I plan to add some more features later. Closes #45 and eliminates the need for #318. If there are any obvious problems with my code that I didn't spot, @ me on Discord or fix them yourself if you want.
Screenshots:
image
image
image


FenyLabs:
You might want to remove config.ts, that was for a feature I haven't implemented yet and isn't used. However, it probably won't show up in the bundle, so it wouldn't cause too many problems if left in.

src/plugins/performance-info/PerformanceView.tsx
```diff
@@ -0,0 +1,97 @@
...
+                this.props.controller().getTimingData().graphAllChanges
+              )
+            }
+            ms
+          </li>
+          <li>
+            <strong>Other: </strong>
+            {() => {
+              const timingData = this.props.controller().getTimingData();
+              return Math.round(
```
jared-hughes, line 58: To be more future-proof, subtract from the total instead of adding up for the "other"

jared-hughes (changes requested):

src/plugins/performance-info/PerformanceView.tsx
FenyLabs: Done! Apparently I forgot subtraction existed when I wrote that code.

jared-hughes (approved):

PR #452, opened 2022-11-23
Refactor WakaTime plugin

### LINKED ISSUE #450: Create a toggle for WakaTime heartbeats: "Desmos Projects" mode.
https://github.com/DesModder/DesModder/blob/4be0b07435c94cf577e978a7a1f12a3b87da94df/src/plugins/wakatime/heartbeat.ts#L23

Allow the user to switch between "Desmos Projects" and `graphName` for sending heartbeats (`graphName` becomes `branch` instead).

### LINKED ISSUE #451: Preload fails if no WakaTime config exists
When the optional chaining operator is used, undefined still !== ""
https://github.com/DesModder/DesModder/blob/5049d016ca82258f678131a936366947ded19313/src/preload/content.ts#L21-L23
If no config exists for the WakaTime plugin preloadContent fails :(

FenyLabs:
- Modifies the WakaTime plugin to more closely match other plugins. It now properly follows https://wakatime.com/help/creating-plugin so AFK time will not be counted (heartbeats are only sent as a result of user interaction).
- Adds an option to split each graph into its own project (current behaviour) with the default being to store all graphs as one project ("Desmos Project"). Strongly recommended by Heavenira. (Resolves #450)
- Fixes a few miscellaneous bugs related to the plugin. (Resolves #451)

### DIFF
--- a/localization/en.ftl
+++ b/localization/en.ftl
@@ -143,6 +143,8 @@ wakatime-name = WakaTime
 wakatime-desc = Track your desmos activity on WakaTime.com
 wakatime-opt-secretKey-name = Secret Key
 wakatime-opt-secretKey-desc = API Key used for WakaTime servers
+wakatime-opt-splitProjects-name = Split Projects by Graph
+wakatime-opt-splitProjects-desc = Store each graph as its own project instead of branches of a unified Desmos Project
 
 ## Performance Display
 performance-info-name = Performance Display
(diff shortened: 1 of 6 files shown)


FenyLabs:
pls test my code before you merge, i'm in the desert and can't really. if you don't i'll test wednesday afternoon probably? also feel free to swap the default for my config option if you don't like it

FenyLabs:
Kind of tested it actually. Option works, heartbeats work, preload is fixed

PR #488, opened 2023-01-31
GLesmos outlines: Version 1

Fasteroid:
## GLesmos Improvements
This PR brings antialiased JFA outlines to GLesmos implicit equations, along with some major refactors and code re-organization.
This should be ready to merge, but I'd like someone else to look over and test it before that happens.

glpreview

### Implemented Features ✅
⭐ Antialiased lines via Jump-Flood algorithm, `O( log n )` time complexity based on screen size.
⭐ GLesmos outline widths are 1:1 with Desmos outline widths.
⭐ GLesmos outlines support color and alpha.
⭐ GLesmos outlines and JFA step are not rendered if disabled or zero width.

### Known Issues :bug:
~~❌ Some implicit outlines develop subtle "blobs" at specific critical points.~~
~~❌ Implicits using an equals sign are filled incorrectly and do not get the "Render with GLesmos" option.~~

### Code Refactors 🔍
➡️ Moved the GLSL stuff into its own file.
➡️ Made a few things more type-safe.
➡️ Improved the shader compilation step.
➡️ Added some comments.
<br>
This project has been over a month in-the-making and I'm super excited to finally present it.
I can't wait to get this out to the community!
<br>


jared-hughes:
I removed a TODO document in favor of GitHub's issue tracker and this PR.

TODOs remaining for this PR, transferred from that document:

- [ ] fix blobby lines bug
- [ ] use less division in the fragment shaders
- [ ] use uniforms for colors, alpha and width
- [ ] outline: Increase width and opacity on equation selected
- [ ] fill: increase opacity on equation selected

The only TODO that would block merge is the first one (blobby lines bug) depending on severity. **Could you share some examples of graphs where you encounter the blobby lines bug?**

(I did not move "GLesmos polygon renderer" to the issue tracker because polygon fills are already optimized draw calls. Feel free to create a new issue arguing your point if you disagree.)

src/plugins/GLesmos/shaders.ts
```diff
@@ -0,0 +1,423 @@
...
+}
+
+const shaderCache = new Map<string, GLesmosProgram>();
+function getShaderProgram(
+  gl: WebGL2RenderingContext,
+  id: string,
+  vertexSource: string,
+  fragmentSource: string
+) {
+  const key = vertexSource + fragmentSource; // TODO: this is terrible
```
jared-hughes, line 99: I'm not sure what's going on here. Why is this terrible? Do you have an idea for an improvement?

src/plugins/GLesmos/shaders.ts
```diff
@@ -0,0 +1,423 @@
...
+    void main(){
+      vec2 mathCoord = texCoord * graphSize + graphCorner;
+      float v = _f0( mathCoord.x, mathCoord.y );
+      outColor = vec4(v, 0, 0, 1);
+    }
+  `;
+
+  const shader = getShaderProgram(gl, id, VERTEX_SHADER, source);
+
+  // TODO: set some uniforms here
```
jared-hughes, line 203: Is this needed?

src/plugins/GLesmos/glesmosCanvas.ts
```diff
@@ -155,72 +104,163 @@ export function initGLesmosCanvas() {
...
-      return {
-        fragmentSource: GLESMOS_FRAGMENT_SHADER.replace(
-          /\/\/REPLACE_WITH_GLESMOS[\s\S]*\/\/REPLACE_WITH_GLESMOS_END/g,
-          shaderCode
-        ),
-        vertexSource: VERTEX_SHADER,
-      };
-    });
+    if (chunk.line_width === 0) {
+      // TODO: this is bad, globals are bad
```
jared-hughes, line 156: Yeah this whole file has been over-globaled from the beginning. A whole bunch of closured variables with a bunch of functions. Perfect use-case for a class; a refactor would be reasonable after the PR is merged.

src/plugins/GLesmos/exportAsGLesmos.ts
```diff
@@ -18,24 +24,29 @@ export function compileGLesmos(
...
     return {
       deps: functionDeps.map(getDefinition),
-      defs: [`${type} ${f}(float x, float y) {\n${source}\n}`],
-      bodies: [
-        `if (${f}(x,y) > 0.0) {` +
-          `  outColor = mixColor(outColor, ${colorVec4(color, fillOpacity)});` +
-          `}`,
+      chunks: [
+        {
+          def: `${type} _f0(float x, float y) {\n    ${source}\n}`,
```
jared-hughes, line 45: Looks like you removed shader combining. For example, `sin(x) < [0,0.1...1]` compiles into 11 different shaders instead of just 1. This might be ok (the shader combining might have been premature optimization), but it would require testing to make sure we don't get a huge performance regression on lists of functions.

Fasteroid: No need to test, I already did-there is a performance hit on lists of functions now.

That said, I'm not sure if it's possible to combine everything into one shader now with the way outlines and fills need to be layered.

Will be actively thinking about this.
Maybe we could revert to old (fast) shader behavior if outlines are off?

src/plugins/GLesmos/shaders.ts
Fasteroid: Using huge strings as map keys just feels bad.  Hashing the entire shader just to place it in a map can't be _that_ cheap.

That said, adding `vertexShader` probably isn't needed since the same one is used everywhere.

src/plugins/GLesmos/shaders.ts
Fasteroid: Since `getShaderProgram` returns an already-built shader if nothing has changed, it would be ideal to offload as many changes as possible into uniforms so we don't compile as much.

This comment is just here to remind someone (me) to do that later, and that it should be done here.  It would likely be a separate PR that packs all function variables into uniforms or something.

src/plugins/GLesmos/glesmosCanvas.ts
Fasteroid: This is part of the reason why I created the `glesmosShaderProgram` class :)

Fasteroid:
> I removed a TODO document in favor of GitHub's issue tracker and this PR.

Awesome!

> The only TODO that would block merge is the first one (blobby lines bug) depending on severity. **Could you share some examples of graphs where you encounter the blobby lines bug?**

I'm pretty sure the implicit in [this](https://www.desmos.com/calculator/ehtiauon9i) graph does it, though I'll be posting an image later just to highlight exactly what I'm referencing.

The issue probably isn't _super important_ though because when it does happen (rare and sparse), the blobbiness equates to only an extra pixel in width around some seed points.

My suspicion is that these points are caused by the derivative of the implicit not being stable around them, since that *is* used in the antialiasing calculations.

Needs more testing to pin down exactly what it is.  I'll write back later today.

Fasteroid:
> I'll be posting an image later

Image of outline blobs from [this](https://www.desmos.com/calculator/nveqovtbh1) graph
image

Got any cheap ways to detect function continuity?

Fasteroid:
Update: The blobs do appear to correlate with numerical instability of the function's derivative:
https://www.desmos.com/calculator/u08tic98nf

Fasteroid:
Going to put the remaining issues on my fork so we can track them...

src/plugins/GLesmos/shaders.ts
```diff
@@ -0,0 +1,419 @@
...
+      float dist = LineSDF( seed * vec4(warp,warp), texCoord * warp ) * max(iResolution.x, iResolution.y);
+
+      float alpha = smoothstep(0.0, 1.0, clamp( dist - float(${chunk.line_width}) * 0.5 + 0.5, 0.0, 1.0 ));
+      outColor = mixColor(outColor, ${chunk.line_color} * vec4(1.0,1.0,1.0,1.0 - alpha));
+    }
+  `;
+
+  const shader = getShaderProgram(gl, id, VERTEX_SHADER, source);
+  gl.useProgram(shader);
+  setUniform(gl, shader, "iDoOutlines", "1i", chunk.line_width > 0 ? 1 : 0);
```
jared-hughes, line 416: Why is `iDoOutlines` a uniform instead of a constant? Probably doesn't really affect performance, but could affect shader compilation time.

Fasteroid: This uniform lets us avoid compiling a separate more performant shader (or copy-pasting lots of code) for when outlines are off.

Fasteroid:
https://github.com/Fasteroid/DesModder/issues/2 is as resolved as it's going to get.  Ready to merge this?

jared-hughes:
The arbitrary 2048 character limit is concerning because it's so low. For example, the fast graph <https://www.desmos.com/calculator/cwol27id84> is at almost 5000 characters. I'd rather have the limit removed to restore to the behavior before this PR.

Relatedly, why did you add the limit with the note "seriously this is for your own good?" I'm guessing it's from some graph that froze your GPU (could you share it?). After this PR, we could probably find some cleaner way to avoid freezing.

Fasteroid:
> The arbitrary 2048 character limit is concerning because it's so low. For example, the fast graph <https://www.desmos.com/calculator/cwol27id84> is at almost 5000 characters. I'd rather have the limit removed to restore to the behavior before this PR.

Reverted, I agree it was probably too aggressive of a limit. My guess is that the graph mentioned only lags because of the nested factorials, which are probably really expensive to compile.

Fasteroid:
It seems the graph you gave has some problems with compiling derivatives for the outlines:
https://www.desmos.com/calculator/cwol27id84

image

My guess is that `emitChunkGL` might not be entirely respecting that certain variables need to be defined before they're used.

Fasteroid:
🎆 🥳 🎆

Issue #494, opened 2023-02-09
Outdated replacements, DesModder currently fails to load entirely!


Fasteroid:
These replacements are now outdated and cause DesModder to fail entirely:
https://github.com/DesModder/DesModder/blob/917b66327be06c0ceb11be3962ba4b9dfb7495b3/src/preload/moduleOverrides/glesmos.replacements#L72-L99

Hotfix PR incoming fast.

Issue #480, opened 2023-01-13
Graceful failures and easy disabling


jared-hughes:
For any detected problem, provide the user a simple way to disable the (estimated) causes of the problem, or disable DesModder entirely. For example, if replacements fails for hide-errors, glesmos, and core, show

```
DesModder (version a.b.c) may have encountered issues in some plugins:
[ ] disable Hide Errors (2 issues detected)
[ ] disable GLesmos (1 issue detected)
[ ] toggle all plugins
Or
[ ] disable all of DesModder (1 core issue detected)
(button:) [apply; greyed out if there's no changes] [reload; greyed out if no changes have been applied]
```

This should be a pretty visible pop-up. Since the extension might still be usable, there should be a button to hide the errors, and put a new button somewhere to show the pop-up again.

Problems can be detected in at least a few ways:
- replacements that match less than once or more than once
- errors in replacements
- actual JS errors; need to be careful here with callbacks and async exprs not sending the errors to the (sync) try-catch
- the main calculator not loading, e.g. loading circle still pulsing after 10 seconds

PR #495, opened 2023-02-09
tiny hotfix

Fasteroid:
Patches the outdated replacements in https://github.com/DesModder/DesModder/issues/494 and makes DesModder load again.

This should probably be merged ASAP to restore partial functionality.
We can investigate what (if anything) this PR breaks later.


jared-hughes:
LGTM. Will probably break #488, but it seems this PR correctly fixes GLesmos.

The current release does not break as horribly as the dev branch. I'll need to fix the brittle issue. Currently, my plan is #480, which I'm in progress of implementing.

Issue #458, opened 2022-12-01
Upgraded evaluation footer


jared-hughes:
Plugin idea based around upgrading the footer:

![image](https://user-images.githubusercontent.com/20214911/205178218-520aadda-0f5a-4c43-bada-136738ab8f9c.png)

- show `L=10  element list of <X>` for lists of points, lists of polygons, etc.
- allow previewing lists by clicking an icon (👁 maybe)
- show `A=action` for actions
- allow previewing actions by clicking an icon
- allow viewing the full value of numbers (all 16 decimals, and NaN vs ±Infinity) by default or by clicking an icon

All this data is already passed back from the worker, it's just not displayed completely

jared-hughes:
Closed by #509

PR #509, opened 2023-02-16
Better Evaluation View (Attempt 2)

FenyLabs:
Better Evaluation View but I'm actually kinda motivated this time

- [x] Show &infin; and -&infin; instead of "undefined"
- [x] Show "NaN" instead of "undefined"
- [x] Show elements inside of lists
- [x] Show colors as rgb values
- [ ] Show points and lists of points as coordinates
- [ ] Allow individual expression toggling


FenyLabs:
Just realised I left the plugin as second in the list. It's one of the few plugins with config that might need to be changed often, but feel free to move it if you want.

FenyLabs:
Closes #458

PR #516, opened 2023-02-28
GL perf improvements

Fasteroid:
When outlines are disabled for an entire group of list implicits (whether it be via 0 width, 0 alpha, or the toggle switch), GLesmos will now use a faster single-pass fill shader.

### DIFF
--- a/src/plugins/GLesmos/Controller.ts
+++ b/src/plugins/GLesmos/Controller.ts
@@ -1,6 +1,6 @@
 import ViewportTransforms from "./ViewportTransforms";
 import { initGLesmosCanvas, GLesmosCanvas } from "./glesmosCanvas";
-import { GLesmosShaderPackage } from "./shaders";
+import { glesmosError, GLesmosShaderPackage } from "./shaders";
 import { Calc } from "globals/window";
 
 export default class Controller {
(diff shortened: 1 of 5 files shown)


Fasteroid:
Last linting issue is here:
 ```ts
  const mains = chunks.map((chunk, id) => {
    return `float f_xy_${id}(float x, float y){
      ${chunk.main}
    }`;
  }).join("\n");

  const colorCalls = chunks.map((chunk, id) => {
    return `if( f_xy_${id}( mathCoord.x, mathCoord.y ) > 0.0 ){
      outColor = mixColor(outColor, ${chunk.color});
    }`;
  }).join("\n");
```
Prettier wants to expand it by 4 lines look like this:
```ts
  const mains = chunks
    .map((chunk, id) => {
      return `float f_xy_${id}(float x, float y){
      ${chunk.main}
    }`;
    })
    .join("\n");

  const colorCalls = chunks
    .map((chunk, id) => {
      return `if( f_xy_${id}( mathCoord.x, mathCoord.y ) > 0.0 ){
      outColor = mixColor(outColor, ${chunk.color});
    }`;
    })
    .join("\n");
```
I don't like this for two reasons—first is that `.map` is no longer visually attached to `chunks`, and second is that the closing curly braces at are now at the same indentation level.  I hope we can agree to disagree on this one.

FenyLabs:
This indented style is common for chained methods, and it's generally best practices to let prettier have its way even when it's not exactly how you like it to look.

FenyLabs:
If you were to ignore it for linting, use a `// prettier-ignore` comment instead of ignoring the entire file.

Fasteroid:
> If you were to ignore it for linting, use a `// prettier-ignore` comment instead of ignoring the entire file.

cool, I'll try this

PR #551, opened 2023-04-13
Fix overflow in DesModder menu #544

### LINKED ISSUE #544: Fix overflow in DesModder menu
E.g. in French:

image

Also FPS in video-creator export.

image

RannyBergamotte:
This fixes #544 . I searched for the line, literally stared at my screen and the inspector for a minute, and then moved two lines, before realizing i moved them to far, and then moved it to where it is now. Needless to say, it works, and I have no idea about code quaility (well just blindly moving two lines), but seems to work fine!

That being said, if I revisit this, and being extremely nitpicky of myself, should the checkbox be at the same height as the first line? If you think so, I can go back and fix it again.
image

### DIFF
--- a/src/components/Menu.tsx
+++ b/src/components/Menu.tsx
@@ -178,22 +178,20 @@ export default class Menu extends Component<{
 }
 
 function booleanOption(controller: any, item: any, plugin: any, settings: any) {
+  const toggle = () =>
+    controller.expandedPlugin &&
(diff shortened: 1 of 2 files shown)


jared-hughes:
Looks good aesthetically (centered checkbox is good).

I've made a extra change, so users can once again click the label to toggle the checkbox.

PR #553, opened 2023-04-16
Add spanish translation

SlimRunner:
I just finished reviewing my changes and @PurpleSunsetGames. I also checked all the GUI elements to see if there were issues with overflow.

I only found one which can be previewed in this graph [https://www.desmos.com/calculator/3q7o0kj3ak?lang=es](https://www.desmos.com/calculator/3q7o0kj3ak?lang=es). Here is a picture of the issue
image
However, it seems it is the fact that Desmos doesn't expect the tray to change size which pushes the toggle off the tray. **I tried to make that translation shorter but there is no way to do** so even when using the sorter foreignism "renderizar" in place off "Generar grafica".

### DIFF
--- a/assets/webstore_screenshots/README.md
+++ b/assets/webstore_screenshots/README.md
@@ -7,10 +7,15 @@ Procedure:
 - If it's actually 1920x1200, run `convert filename.png -resize 66.67% filename.png` to downscale to 1280×800
   - bulk script: `for f in *.png; { convert $f -resize 66.67% $f; }`
 
-Graphs:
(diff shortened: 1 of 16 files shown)


PurpleSunsetGames:
You could shorten "Generar Gràfica con GLesmos" to "Usa GLesmos" and the meaning would still be clear. You could also try  using "dibuja" (draw) which, in this context, means pretty much the same thing as render. "Puede activarse con un clic" can also just be shortened to "cliqueable."

jared-hughes:
Is this ready to merge?

Issue #68, opened 2021-05-02
Customize thumbnail image


jared-hughes:
By default, this could create thumbnail images based on the actual viewport size instead of the arbitrary 200×200 default for Desmos.

This could also allow uploading custom thumbnails like u/-fasteroid used in https://www.desmos.com/calculator/desmossets (see https://www.reddit.com/r/desmos/comments/m6ezp8/desmos_set_utilities/).

jared-hughes:
Keep in mind that the image should probably still be 200×200 to avoid increasing the thumbnail file size. Use `targetPixelRatio` as `200 / min (x bound, y bound)`, though these should be customizable.

This plugin can be controlled through a selection (maybe change plugin settings to `string` instead of `boolean` and migrate paste state to strings based on versioning):
- default (200x200, pixel ratio 1)
- 200x200, vmin pixel ratio
- Use first image in the graph as thumbnail (need to limit size + require png. `alert()` in overriden module if these conditions are not met).

jared-hughes:
Use a UI to set the manual photo, and store the data url for it in the graph metadata? Unfortunately increases graph size, so might want to put a checkbox on whether the uploaded image should be stored in the graph data.

PR #555, opened 2023-04-20
Plugin custom save

ronwnor:
lil plugin to save graphs with custom thumbnails
image


.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 dist
 *.zip
 coverage
+calculator.js
```
FenyLabs, line 5: Might want to remove this, probably just for development

src/plugins/custom-save/Controller.ts
```diff
@@ -0,0 +1,22 @@
...
+
+function updateView() {
+  desModderController.updateMenuView();
+}
+
+export default class Controller {
+  checked: Boolean;
+
+  constructor() {
+    this.checked = false;
```
FenyLabs, line 11: Tiny nitpick but you could probably use a default value here

src/plugins/custom-save/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  controller = new Controller();
+  return controller;
+}
+
+const CustomSave: Plugin = {
+  id: "custom-save",
+  onEnable,
+  onDisable: () => {},
+  onConfigChange: () => {
+    console.log("AAAAAAA");
```
FenyLabs, line 16: Again might want to remove before merge

src/preload/moduleOverrides/custom-save.replacements
```diff
@@ -0,0 +1,77 @@
+# Custom Save Replacements
+
+*plugin* `custom-save`
+
+## Add a checkbox to the save dialog
+
+*module* `custom-save`
```
FenyLabs, line 7: Not sure if fireflame cares about this anymore, but I think you're overriding `calculator-shell/save-dialog`

src/preload/moduleReplacements.ts
```diff
@@ -23,6 +24,7 @@ const replacementStrings = [
   debugMode,
   extraExpressionButtons,
   showTips,
+  CustomSave,
```
FenyLabs, line 27: camelCase instead of PascalCase for consistency?

src/plugins/custom-save/Controller.ts
```diff
@@ -0,0 +1,22 @@
+import { desModderController } from "script";
+
+function updateView() {
+  desModderController.updateMenuView();
```
FenyLabs, line 4: Is this doing anything? Was under the impression this was to update DesModder menus/pillboxes, not Desmos's. Perhaps Calc.controller.updateViews()?

FenyLabs (changes requested):
Just a couple tiny things to change to clean up. However, the plugin feels a bit awkward to use and depends on video creator output being a square. Not sure how to solve this, but perhaps the checkbox could be moved to be a part of video creator ui? or it could somehow be made more clear how to use the plugin. Currently it's not very clear that the rename modal also contains the checkbox, so perhaps that dropdown element could be renamed or something.

src/plugins/custom-save/index.ts
ronwnor: ok this is embarrassing ///

.gitignore
ronwnor: oh whoops

src/preload/moduleOverrides/custom-save.replacements
jared-hughes: It's ok. The specific module names are pretty irrelevant now. They can probably be deleted throughout, but they're not doing much harm.

jared-hughes:
It's OK if the video creator output is not a square, as long as the thumbnail gets cropped to be square.

Another consideration to the screenshot dimensions is if they're too big. Maybe downscale the image by default to a 400x400 max? Or maybe the best approach could be to warn the user if they're trying to upload a huge image, and give them an option to cancel.

For dynamic resize in JS (tweak the drawImage call).

```js
const canvas = document.createElement("canvas");
const ctx = canvas.getContext("2d");
ctx.drawImage(img, 0, 0, 300, 300);
const dataurl = canvas.toDataURL("image/png");
```

I almost never go into the "rename" menu, so I resonate with Feny's idea of moving the button to the video creator menu. It's easier to add a confirmation button before activating the native save-graph menu than to create a bunch of replacements.

jared-hughes:
This would fix #68.

jared-hughes:
Closing this because stale. If you want to restart it, I suggest going from main and re-applying changes.

Issue #582, opened 2023-05-17
Function Exponent handling


jared-hughes:
The FunctionExponent in Raw is currently converted to exponent of a function in Aug. E.g. `f(x)^2` goes to `(f(x))^2`. This causes a problem when it's actually a product, e.g. `x(x+1)^2` converts to `(x*(x+1))^2`. This screws up the cubic bezier helper `C(p)` in [Owl](https://www.desmos.com/calculator/t2e1wyoc7l).

A fix could involve adding FunctionExponent to both Aug and TextAST.

A more elegant solution may be simply changing the Latex emitter to reduce syntax when doing an exponent of a function application. Given a Text input of `f(x+1)^2`, instead of emitting `(f((x)+(1)))^2`, emit `f((x)+(1))^2`. This works whether `f` is a function or non-function.

jared-hughes:
Fixed in #599

PR #599, opened 2023-05-22
Reduce parens emitted to raw

jared-hughes:
Adds and uses `augNeedsParens` to determine if parentheses are needed on the emit of an expression to raw. The previous behavior was effectively always returning true. Now, we convert many cases to false, making ronwnor slightly happier.

Also download calculator_desktop before tests, to be able to test against Desmos's parser.


jared-hughes:
Fixes #582

PR #600, opened 2023-05-26
zh-CN translations

StaticReasons:
zh-CN translations; .gitignore file added .vscode directory

### DIFF
--- a/localization/zh-CN.ftl
+++ b/localization/zh-CN.ftl
@@ -1,0 +1,179 @@
+# File Conventions:
+# Everything related to a plugin starts with the ID of the plugin
+# Hardcoded in the Typescript:
+# [pluginID]-name = Name
+# [pluginID]-desc = Description
+# [pluginID]-opt-[optionKey]-name = Option Name
+# [pluginID]-opt-[optionKey]-desc = Option Description
+
+## General
+menu-learn-more = 了解更多
+menu-desmodder-plugins = DesModder 插件列表
+menu-desmodder-tooltip = DesModder 菜单
+
+## Category names
+category-core-name = 核心功能
+category-utility-name = 小工具
+category-visual-name = 界面设置
+category-integrations-name = 对外集成
+
+## GLesmos
+GLesmos-name = GLesmos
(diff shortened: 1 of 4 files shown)


localization/zh-CN.ftl
```diff
@@ -0,0 +1,179 @@
...
+
+## Category names
+category-core-name = 核心功能
+category-utility-name = 小工具
+category-visual-name = 界面设置
+category-integrations-name = 对外集成
+
+## GLesmos
+GLesmos-name = GLesmos
+GLesmos-desc = 使用GPU渲染隐函数等图象。刷新页面后此功能将自动关闭。极少数情况下此功能可能会导致UI运行卡顿或卡死; 若出现问题请刷新网页重试。
```
jared-hughes, line 22: Google Translate suggests "GPU" could be "显卡". I have no way to verify this; feel free to make the edit if it seems right to you.

localization/zh-CN.ftl
```diff
@@ -0,0 +1,179 @@
...
+# Missing: error messages
+
+## Tips
+show-tips-name = 显示小贴士
+show-tips-desc = 在表达式列表的底部显示小贴士
+# Missing: all tips. Is it worthwhile?
+
+## Text Mode
+text-mode-name = 文本模式 BETA
+text-mode-desc = 可能会出现一些bug。临时的使用文档请参见：
```
jared-hughes, line 35: Maybe "bug" could be translated like "problem"?

jared-hughes (commented):
This is a great addition, thanks!

I left two comments regarding words that you left as English but might be able to be translated. It's up to you though. If you commonly see them untranslated, feel free to keep them the same as they are now (but let me know so I can merge).

StaticReasons:
Thanks, and your comments have noticed me to pay more attention on those kept-in-English words.
From my personal sense, some other words may be better translated, and it's OK to keep the word "GPU". And perhaps that's all.

StaticReasons:
However I have a small problem: the suggestion of "npm run audit-lang <lang>" seemed doesn't work. I used node@18.7.0 and npm@8.15.0, and the error output is like this:
Screenshot from 2023-05-27 23-20-48
So I didn't use it to check my translation.

jared-hughes:
The audit-langs is not a problem. I fixed it in https://github.com/DesModder/DesModder/commit/f1ea88bb0b8e7fea89d0f4a904bcc3776fb0d000, so I could run it manually. Looks good; I'll go ahead and merge.

PR #603, opened 2023-05-31
Fix replacements

FenyLabs:
Draft PR, extension currently works? (except for BEV) but goes crazy panic mode


jared-hughes:
Thanks for the kickstart!

PR #621, opened 2023-07-02
Distinguish NaN, +∞, -∞ in lists.

jared-hughes:
This code was deleted a while ago in mistake (it failed to replace successfully, and I thought it wasn't needed because I didn't check lists).

Props to @FenyLabs for spotting the mistake.


FenyLabs:
Thanks fireflame

PR #652, opened 2023-07-10
Update spanish translation jul23

SlimRunner:
I think it is good right now, but I want to review it another time tomorrow (well today lol) in case there is something I'm not seeing that automatic proofing may not have caught. Also, I will ask in Discord if a second pair of eyes can give it a look as well.

I'll comment with a thumbs up once that's done so you know when to merge.


SlimRunner:
I already gave it a few extra proof reading passes. It seems alright to merge.

SlimRunner:
Also, sorry about that force push... I was trying to squash a commit I had already pushed. I did +branchname instead of --force (). I'm not entirely sure that makes it any better, or whether this is bad.

Let me know if you prefer any force push, so I'll avoid it in the future.

jared-hughes:
Force pushing to feature branches is often okay (since they're short-lived, and typically other people don't need to push changes to your feature branch). One thing it can break is existing review comments, which are linked to specific commits.

PR #675, opened 2023-07-26
Update Spanish translations

SlimRunner:
Included the latest plugin into the spanish translations. This is already proof-read.


jared-hughes:
Nice! Thanks for being on top of this.

PR #695, opened 2023-08-14
Enable Text Mode Beta in Geometry Beta

jared-hughes:
To see changes: Open https://desmos.com/geometry-beta post-PR. Text Mode can then be enabled. When inside Text Mode, inserting and deleting geo objects from the graphpaper are reflected in the text representation.

This also lays the groundwork for passing incremental changes, by using the `mapIDStmt` from the last parse to match up IDs.

Known bug:
- if the `**dcg-geo-folder**` is empty and a geo object is inserted through the toolbar, it doesn't get indented. Not worth the effort.

### DIFF
--- a/src/MainController.ts
+++ b/src/MainController.ts
@@ -28,7 +28,6 @@ export default class DSM extends TransparentPlugins {
     super();
     // default values
     this.forceDisabled = window.DesModderPreload!.pluginsForceDisabled;
-    if (Calc.controller.isGeometry()) this.forceDisabled.add("text-mode");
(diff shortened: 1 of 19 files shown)


radian628 (approved):
This looks good to me for the most part! I have to admit I'm not super familiar with CodeMirror and I don't have all that much experience with the Geometry Calc, so I don't think what I've covered is anywhere near exhaustive. But I think I've at least ruled out low-hanging fruit like spelling errors and commented-out code.

jared-hughes:
Thanks for the review!

PR #696, opened 2023-08-15
Show IDs in Text Mode with Debug Mode enabled

### LINKED ISSUE #693: Text Mode Debug Mode
Simple: When Debug Mode is enabled, show expression IDs instead of indices in Text Mode.

jared-hughes:
There might currently be no way to get a difference between IDs and indices while in text mode. 
But merging with https://github.com/DesModder/DesModder/pull/695 allows for actually testing this code:

image

Fixes #693

### DIFF
--- a/src/plugins/debug-mode/index.ts
+++ b/src/plugins/debug-mode/index.ts
@@ -9,9 +9,11 @@ export default class DebugMode extends PluginController {
     // The displayed indexes are stored in some state somewhere, so
     // update the state first before updating views
     Calc.controller.updateTheComputedWorld();
+    this.dsm.textMode?.updateDebugMode();
   }
 
   afterDisable() {
     Calc.controller.updateTheComputedWorld();
+    this.dsm.textMode?.updateDebugMode();
   }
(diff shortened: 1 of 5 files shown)


radian628:
Sorry for the late review! I just tried to test this on the graphing calc, but I can't seem to get the IDs to show up in Text Mode even though Debug Mode is enabled. Text Mode also seems to automatically switch the IDs to an increasing sequence of integers when one switches out of it. Not sure if this is intentional and it's doing that because those _are_ the IDs now.

radian628:
For context, I'm on Windows 11 and Microsoft Edge.

radian628:
Oh wait, is this specifically designed to work in combination with #695? I might be misunderstanding.

jared-hughes:
Sorry this was confusing. Yes, Text Mode defaults to the IDs being equal to the index. The only IDs it keeps are those which are not simply digits. E.g. it preserves `**dcg-geo-folder**`. So the constructions folder gets preserved through Text Mode:
image

PR #697, opened 2023-08-15
facets

jared-hughes:
- Track enabled plugins with Codemirror State
- Track plugin settings with CM State
- Convert pillbox-menus to a CM ViewPlugin
- Migrate performance-view and video-creator to cm
- Migrate builtin-settings to CM
- Migrate bunch of plugins to CM
- Migrate expr-action-buttons (and friends) to CM
- Migrate Text Mode to CM
- Migrate metadata to CM
- Rename cmPlugins to plugins


jared-hughes:
oops

PR #706, opened 2023-08-28
Add emit config to remove spaces and newlines

jared-hughes:
- `astToText` and related functions now take `TextEmitOptions`, which allows disabling spaces and newlines
- Adds a "Format" button that also lets you toggle "Spaces" and "Newlines" (If spaces is TRUE, then newlines is forced to TRUE)
   image
- Places footers at the end of lines, so they don't interject into the middle of lines
   image


radian628 (changes requested):
Looks good to me! All I'd suggest is having some kind of tooltip that explains what "Spaces" and "Newlines" mean. (I assume "Spaces" toggles indentation and space between things like operators; and "Newlines" toggles whether to put everything in one line or not)

PR #707, opened 2023-08-28
Migrate most baseurl-relative imports to #-imports

jared-hughes:
- Remove sort-imports plugin
- Convert a bunch of imports relative to baseurl (like `import "plugins"`) to directly-specified imports

Main advantage here is that everything is explicitly specified, which'll make the project easier to reason about (including what is going on with text-mode-core).

Incidentally, text-mode-core is also the main problem with entirely getting rid of `baseurl: "src"` in the tsconfig. In the way it's currently structured, it can't import stuff outside of the text-mode-core folder without going relative to the baseurl. Maybe tsconfig `paths` + some esbuild setting could fix this, but that's a job for a later PR.


radian628 (commented):
Looks all good to me!

PR #713, opened 2023-09-02
Custom MQ Config Fixes

Naitronbot:
### Remove nu from More Greek Letters
(sorry fireflame i swear this was an accident)

### Skip over static MathQuill fields
Skips over static MathQuill fields when applying greek letter injections. Caused error to be thrown when a table was present in the expression list.


jared-hughes (approved):
Nice

PR #714, opened 2023-09-04
(WIP - will need to get back up to date) Setup Text Mode Editor for npm + Clean out 'Calc'

jared-hughes:
Many changes:
- rearranging some directories to make dependencies clear:
  - /text-mode-editor is its own directory. The /src/plugins/text-mode just deals with mounting and toggling the editor inside Desmos.
  - /common contains DSM metadata, DCGView, common components, utils, and global helpers
  - /tests contains test helpers
- Each of these is marked as a TS subproject (has its own `.tsconfig` which extends the base tsconfig, and another tsconfig references it (`references: [{path: ...}]`)), and some are marked as npm packages (has its own `package.json`). They can't be built individually, but (I hope) this is a good structure for the future. These can be imported by `file:` inside dependencies inside package.json.
- Enable eslint rule import/no-extraneous-dependencies, to ensure each package doesn't import stuff not declared in its package.json.
- Disable eslint rule rulesdir/no-external-imports. It got in the way more than it was helpful.
- Pass a specific `calc` around everywhere. Now, `Calc` is only used in three places (grep for `window.Calc` for them, since `Calc` is used as the type of calc now). In theory, this allows DesModder to run on any page with one more calculator instances, but replacements and content script are the blocking points for now.
  1. In `src/script.ts`, the main initialization, `const dsm = new DSM(window.Calc);`
  2. In GLesmos replacements, `.drawGLesmosSketchToCtx?.(window.Calc.controller, ...` (I didn't bother to figure out how extract a Desmos controller from that point in code)
  3. In `src/preload/script.ts`, where it waits to inject the main script until `window.Calc` is loaded.
- Expose TextModeEditor as an API

Some changes that *didn't* happen:
- exposing any other DesModder plugins as an API
- fixing the bundling situation (text-mode-core and moo and codemirror all common utilities are currently bundled into one big index.js when publishing text-mode-editor).
- emitting types from the packages
- Delete the unused `attach` and `propGetSet`

Observations: 
- wakatime and video-creator both rely on `calc._calc.globalHotkeys`, which only exists on desmos.com. This is for `getCurrentGraphTitle`. API instances don't have a graph title. Would need some more graceful fallback.
- set-primary-color favicon changes rely on desmos.com favicons. Shouldn't make any change on other sites


jared-hughes:
Plan for the second attempt:

One PR:
- Pass a specific `calc` around everywhere

One PR:
- Rearrange dependencies and mark TS/JS subprojects
- play with eslint rules

One PR:
- Expose TextModeEditor API

PR #716, opened 2023-09-11
Don't auto-subscript length-1 identifiers

jared-hughes:
Steps to reproduce:
- Start writing `c(x)=rgb(x,x,x)`
- Before commit: After writing the first comma, the cursor would move after the (ghost) close paren. After commit: the cursor stays where it is


radian628 (approved):
Looks good to me!

Issue #683, opened 2023-08-08
Japanese translation?


radian628:
Given the fact that there's a large Japanese math graph art community, it might be good to create a Japanese translation for DesModder. I wouldn't be able to do this as I do not know Japanese, but I'm sure that quite a few people (e.g. ronwnor) do.

zukizukizuki:
I'm Japanese. I would like to implement this translation, where should I translate?

radian628:
Thank you for the offer to translate! All of the translations and the logic that implements them is [here](https://github.com/DesModder/DesModder/tree/main/localization).

To add the translations, you should create a text file called `ja.ftl` in that directory and add the translations there. DesModder uses [Project Fluent](https://projectfluent.org/) for localizations.

To get DesModder to use the translation, edit [this file](https://github.com/DesModder/DesModder/blob/main/localization/i18n-core.ts) to import `ja.ftl` and call `addLanguage` with it at the bottom (in other words copy the code for the other languages, but use the Japanese localization file and language code instead). 

I have to admit I'm not 100% certain that this is _all_ the steps to get this to work (@jared-hughes was the one who set up the internationalization system) but I'm pretty sure this will work.

radian628:
Somehow I didn't notice the README also has an article on this lol. I'll link it here: https://github.com/DesModder/DesModder/blob/main/docs/DEVELOPMENT.md#example-creating-a-new-translation-file

zukizukizuki:
@radian628 
Thank you😊I'll try it.
I will contact you if I have any more questions.

PR #718, opened 2023-09-11
Japanese translation

zukizukizuki:
## Overview
The following is a translation of the Japanese text.
https://github.com/DesModder/DesModder/blob/main/docs/DEVELOPMENT.md#example-creating-a-new-translation-file

I have translated the items listed in [en.ftl](https://github.com/DesModder/DesModder/blob/main/localization/en.ftl).
I do not translate content that, in my judgment, can be understood as is.

## issues
https://github.com/DesModder/DesModder/issues/683


zukizukizuki:
@radian628
Please check it.

localization/i18n-core.ts
```diff
@@ -52,4 +53,5 @@ function addLanguage(locale: string, ftl: string) {
 addLanguage("en", enFTL);
 addLanguage("es", esFTL);
 addLanguage("fr", frFTL);
+addLanguage("fr", jaFTL);
```
FenyLabs, line 56: You probably want addLanguage("ja")

zukizukizuki: sorry🙇‍♂️ Im fixed it.

localization/ja.ftl
```diff
@@ -0,0 +1,303 @@
...
+## General
+menu-learn-more = 詳細
+menu-desmodder-plugins = DesModderプラグイン
+menu-desmodder-tooltip = DesModderメニュー
+
+## Category names
+category-core-name = カテゴリ名
+category-utility-name = カテゴリ用途名
+category-visual-name = カテゴリ表示名
+category-integrations-name = カテゴリ統合名
```
jared-hughes, line 18: These seem to have translated more than necessary. For example, "カテゴリ統合名" translates back to  "Category integration name", but it should be just "Integrations"

jared-hughes (commented):
I know you said
> I do not translate content that, in my judgment, can be understood as is.

But there are *many* more untranslated strings here than in the other language files. 
In particular, most of the "Video Creator" section is untranslated, and many `-name` translations (even if the corresponding `-desc` is translated).

Is it common to mix Japanese and English text on the same page?

localization/ja.ftl
zukizukizuki: Thank you for your comment. Fixed.

localization/ja.ftl
```diff
@@ -0,0 +1,303 @@
...
+wolfram2desmos-opt-reciprocalExponents2Surds-name = ラジカル記法
+wolfram2desmos-opt-reciprocalExponents2Surds-desc = 1未満の小数の累乗をラジカル等価（surd）に変換する。
+wolfram2desmos-opt-derivativeLoopLimit-name = デリバティブの拡大
+wolfram2desmos-opt-derivativeLoopLimit-desc = ライプニッツ記法のn階微分を反復微分に展開する。（10回まで）
+
+## Pin Expressions
+pin-expressions-name = pinの表現
+pin-expressions-desc = リスト編集モードからのpin式
+pin-expressions-pin = Pin
+pin-expressions-unpin = Unpin
```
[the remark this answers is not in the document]
zukizukizuki: Thank you for your comment. Fixed.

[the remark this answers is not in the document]
ronwnor: 「固定」perhaps? I don't think there's a good verb for "unpin" though. maybe 「外す」.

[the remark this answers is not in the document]
zukizukizuki: It was certainly easier to understand that way. Fixed.

zukizukizuki:
> But there are _many_ more untranslated strings here than in the other language files. In particular, most of the "Video Creator" section is untranslated, and many `-name` translations (even if the corresponding `-desc` is translated).
> 
> Is it common to mix Japanese and English text on the same page?

The parts that I felt did not need translation were the parts where the meaning did not change between Japanese "katakana" and English letters, but it is true that too many English letters are unfriendly, so I modified the translation to "katakana".

zukizukizuki:
> But there are _many_ more untranslated strings here than in the other language files. In particular, most of the "Video Creator" section is untranslated, and many `-name` translations (even if the corresponding `-desc` is translated).
> 
> Is it common to mix Japanese and English text on the same page?

@jared-hughes 
The parts that I felt did not need translation were the parts where the meaning did not change between Japanese "katakana" and English letters, but it is true that too many English letters are unfriendly, so I modified the translation to "katakana".

jared-hughes (approved):
Thanks for these translations!

PR #720, opened 2023-09-11
Add short message and splash in console

### LINKED ISSUE #607: Show message in console clarifying calculator_desktop.js "error"
I've noticed a lot of confusion for quite some time about the warn that calculator_desktop.js "failed to load" due to desmodder blocking it. Might be worth adding a message in console clarifying that.

FenyLabs:
Extremely basic PR, closes #607

### DIFF
--- a/src/preload/script.ts
+++ b/src/preload/script.ts
@@ -43,7 +43,7 @@ async function load(pluginsForceDisabled: Set<string>) {
   Console.warn(
     BROWSER === "firefox"
       ? `%cThe warning above (Loading failed for the <script> with source...) is intentional and does not indicate a bug.`
-      : `%cThe error above (GET https://... net::ERR_BLOCKED_BY_CLIENT) is intentional and does not indicate a bug.`,
+      : `%cThe error above (net::ERR_BLOCKED_BY_CLIENT) is intentional and does not indicate a bug.`,
     "font-weight: bold;"
   );
 
--- a/src/preload/script.ts
+++ b/src/preload/script.ts
@@ -1,8 +1,4 @@
-import { addForceDisabled } from "../panic/panic";
-import moduleReplacements from "./moduleReplacements";
-import { insertElement, replaceElement } from "./replaceElement";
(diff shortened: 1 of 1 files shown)


src/preload/script.ts
```diff
@@ -40,6 +40,18 @@ function getCalcDesktopURL() {
 async function load(pluginsForceDisabled: Set<string>) {
   if (window.location.href.includes("geometry-legacy")) return;
 
+  Console.warn(
+    BROWSER === "firefox"
+      ? `%cThe warning above (Loading failed for the <script> with source...) is intentional and does not indicate a bug.`
+      : `%cThe error above (GET https://... net::ERR_BLOCKED_BY_CLIENT) is intentional and does not indicate a bug.`,
```
jared-hughes, line 46:
```suggestion
      : `%cThe error above (net::ERR_BLOCKED_BY_CLIENT) is intentional and does not indicate a bug.`,
```
Since it doesn't match for me currently.
image

src/preload/script.ts
```diff
@@ -1,15 +1,15 @@
...
 import {
   postMessageUp,
   listenToMessageDown,
   arrayToSet,
 } from "#utils/messages.ts";
 import { pollForValue } from "#utils/utils.ts";
+import { addForceDisabled } from "../panic/panic";
+import moduleReplacements from "./moduleReplacements";
+import { insertElement, replaceElement } from "./replaceElement";
+import { fullReplacementCached } from "./replacementHelpers/cacheReplacement";
```
jared-hughes, line 12: Looks like these imports got auto-sorted? I took out prettier-plugin-sort-imports because it failed with some update and/or caused confusion due to the `#` introduction (we might want to add it back at some time, but I haven't seen it super helpful). An `npm clean-install` might remove it.

jared-hughes (changes requested):
Thanks for this!

src/preload/script.ts
FenyLabs: I was a bit confused about this too. My local prettier did that and it passed linter, so I just left it.

jared-hughes (approved):

PR #725, opened 2023-09-20
Add custom-mathquill-config option to reduce spacing around "f"

FenyLabs:
The default spacing felt a bit excessive, this PR adds an option to reduce that

### DIFF
--- a/localization/en.ftl
+++ b/localization/en.ftl
@@ -264,7 +264,7 @@ compact-view-opt-noSeparatingLines-desc = Removes the separating lines between e
 compact-view-opt-highlightAlternatingLines-name = Highlight Alternating Lines
 compact-view-opt-highlightAlternatingLines-desc = Highlights alternating expressions so that they can be easily distinguished from one another.
 compact-view-opt-hideEvaluations-name = Collapse Evaluations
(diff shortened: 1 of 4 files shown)


FenyLabs:
I ended up doing 0em left, 0.05em right, seemed right to me

localization/en.ftl
```diff
@@ -301,3 +301,5 @@ custom-mathquill-config-opt-leftIntoSubscript-name = Left/Right into Subscripts
 custom-mathquill-config-opt-leftIntoSubscript-desc = Moving the cursor left or right will go into subscript instead of superscript
 custom-mathquill-config-opt-extendedGreek-name = More Greek Letters
 custom-mathquill-config-opt-extendedGreek-desc = Enables replacements for all supported greek letters
+custom-mathquill-config-opt-lessFSpacing-name = Less F spacing
```
Naitronbot, line 304: Maybe make this a lowecase f, since it only applies to lowecase?

Naitronbot (approved):

SlimRunner (approved):
It looks good. The " instead of ' aligns with the rest of the options.

Also, oddly enough I would have sworn I made this change when I translated the mathquill options to Spanish because I used double quotes there. I guess it slipped my mind.

jared-hughes (approved):

PR #727, opened 2023-09-22
Fix pillbox buttons

FenyLabs:
Removed a replacement that didn't seem to matter anymore (the container no longer has a predicate) and added a new class responsible for pillbox sizing.

### DIFF
--- a/src/plugins/pillbox-menus/components/PillboxContainer.less
+++ b/src/plugins/pillbox-menus/components/PillboxContainer.less
@@ -1,6 +1,7 @@
 .dsm-pillbox-buttons {
   display: flex;
   flex-direction: column;
+  gap: 5px;
   &.dsm-horizontal-pillbox {
     flex-direction: row-reverse;
   }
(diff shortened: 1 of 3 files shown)


FenyLabs:
Tested on main and geometry calculator, with mobile and desktop layouts, and with zoom buttons disabled.

FenyLabs (commented):
Unit test fail is not my fault, seems to be something missing from the JSDom environment for the text-mode-core test?

jared-hughes:
Yeah don't worry about the unit test.
Looks like the buttons are more squished than they used to be. Maybe some gap/margin property was moved to different elements.
PR:
image
Old:
image

FenyLabs:
Oh huh, didn't notice that

jared-hughes:
Probably changed since you made the PR due to the 3d orientation buttons being placed horizontal.

FenyLabs:
That should hopefully do it

jared-hughes:
Looks good!

PR #740, opened 2023-10-01
update spanish translation

SlimRunner:
I already read proofed this translation


jared-hughes (approved):
Nice! Thanks for the translations upkeep.

PR #764, opened 2023-10-12
Fix calculator/geometry swap in set-primary-color

jared-hughes:
Was ported backwards from the ternary `isGeometry ? [...] : [...]` (which was confusing because "calculator" should be listed first typically, and "geometry" second)


radian628 (approved):
idk how this is gonna break anything so might as well approve lol

PR #771, opened 2023-10-22
Fix pillbox menus in /geometry

jared-hughes:
- Always use vertical pillbox in geometry (this removes a broken replacement)
- Hide "pick a tool to get started" when opening pillbox menu


radian628 (approved):
Can confirm that this works in Firefox!

Issue #754, opened 2023-10-07
Improvements to Video Creator in 3D


jared-hughes:
- Allow specifying spin speed and axis precisely. (As XY and Z-forward Euler angles?) -- advance by this amount between screenshots.
- Default the spin speed to zero.
- Allow specifying an initial rotation.

ronwnor:
asyncScreenshot still seems to capture plots incorrectly. it's a vanilla bug but still worth mentioning
![image](https://github.com/DesModder/DesModder/assets/109063662/9cf4ae99-3f90-4c27-8188-9906ed5ee8f9)
![image](https://github.com/DesModder/DesModder/assets/109063662/0672a17a-08a8-44ea-a620-4fa562dd0689)

jared-hughes:
Thanks for bringing up that issue.

radian628:
Idea: being able to control camera axis and scale using a variable. Perhaps even the same thing with the graphpaper bounds in `/calculator`.

FenyLabs:
If you're going to implement that functionality with variables, why not just transform your shapes within the calculator? Seems a bit weird to have animations that you can't see normally but can see when you start a video creator capture...

radian628:
The main thing is that it'd be less annoying to implement camera moves for existing graphs, especially if you don't have animation in mind for some big project and find later that you want to animate it, but don't have any variables for properly transforming the scene. Not to mention a lot less laggy and a lot less temporal artifacts for stuff like 3D implicits with lots of detail.

On the other hand I do think @jared-hughes's solution is a lot simpler and if we make variable-driven camera controls too flexible they could risk becoming something people use for general camera control.

jared-hughes:
Animations I've seen in the wild just rotate at a constant speed in the (math) xy plane. So it could be as simple as a single number input there.

jared-hughes:
I've found it's intuitive to specify the rotation as two Euler angles. First applied is a rotation xyRot that rotates the (view cube's) x axis towards the y axis. Then a rotation zTip that tips the z axis towards the camera.

One PR should add:

- [ ] show the current orientation, in (zTip,xyRot) form (should respect degree mode) (update periodically if spinning? Or disable spinning at start)
- [ ] (optional) allow setting the current orientation (should respect degree mode)
- [ ] slider capture: allow setting the end orientation (should respect degree mode)
- [ ] simulation capture, ticks capture: allow setting change in orientation per tick. (should respect degree mode)

Related:
[...]

PR #772, opened 2023-10-22
Use sync screenshot for 3d capture

jared-hughes:
Fixes vertical stretch, mentioned in https://github.com/DesModder/DesModder/issues/754#issuecomment-1751646517


radian628 (approved):
Works in FF--- no vertical stretching: 
image

PR #781, opened 2023-11-18
Fix rho

Naitronbot:
Rho can be now typed by default, caused an issue with rho being removed when the plugin is disabled.


jared-hughes (approved):

jared-hughes:
Thanks

PR #820, opened 2024-02-26
Handle operatornames in Intellisense

### LINKED ISSUE #705: Intellisense: Function param help documentation does not handle builtins properly
image

jared-hughes:
Fixes #705. Doesn't show useful documentation, but at least it's not saying `n` for `sin()`.

image
image

### DIFF
--- a/src/plugins/intellisense/latex-parsing.tsx
+++ b/src/plugins/intellisense/latex-parsing.tsx
@@ -86,22 +86,18 @@ function isIdentifierSegment(cursor?: MQCursor): cursor is MQCursor {
 // identifiers are composed of the following structure:
 // (operatorname* | varname) subscript?
 
-function tryGetMathquillIdent(
-  mq: MathQuillField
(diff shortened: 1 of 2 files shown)


src/plugins/intellisense/latex-parsing.tsx
```diff
@@ -204,26 +215,12 @@ export function getPartialFunctionCall(
...
-        ltx2 &&
-        ltx &&
-        isIdentStr(ltx2 + ltx) &&
-        cursor?.[1]?.ctrlSeq === "\\left("
-      ) {
-        const str = latexStringToIdentifierString(ltx2 + ltx);
-        return {
-          ident: str as string,
-          paramIndex,
-        };
```
jared-hughes, line 226: This was handling the case where you need to look at two mq nodes to get a full identifier. I think that's only subscripts like `f_{123}(x)`, but the `rawTryGetMathquillIdent` correctly does that.

PR #831, opened 2024-03-22
Bundle ffmpeg core js and wasm

FenyLabs:
Was trying to update to 0.12.x but there were some weird issues with the script loading. Seems like others have them too and it's 0.12.x specific, so I kept it on 0.11.x.

### DIFF
--- a/esbuild.mjs
+++ b/esbuild.mjs
@@ -89,6 +89,13 @@ const opts = {
     copy({
       resolveFrom: "cwd",
       assets: {
+        from: [`./node_modules/@ffmpeg/core/dist/ffmpeg-core.worker.js`],
+        to: `${outdir}/lib/ffmpeg-core.worker.js`,
+      },
+    }),
+    copy({
+      resolveFrom: "cwd",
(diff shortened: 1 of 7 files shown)


FenyLabs:
Huh, just tested it out and it doesn't seem to be working. Was working earlier, idk what happened. Will mess around and try to fix

FenyLabs:
Should be working now, it was working earlier because of leftover artifacts from esbuild-plugin-copy.

jared-hughes:
The main benefit was providing offline support for loading Video Creator, but offline support is already provided by the browser caching the resources indefinitely https://www.unpkg.com/#cache-behavior. The only problem arose in a combination of factors:
- Using Firefox
- With the Devtools open
- With the network forced "Offline"
The problem does not happen in Firefox with the network disconnected the regular way.

To avoid adding 8.5 MB to the 0.5 MB bundle (of every dsm install), we're deciding not to merge.

Thanks for the PR though. Glad to be confident about what's going on here.

PR #835, opened 2024-03-28
add strong-mismatched-brackets option to plugin custom-mq-config

ronwnor:
making this pr to fix something that bothered me for the longest time; 
brackets and parentheses matching with each other
image

I put it in the custom-mathquill-config plugin because I think it fits there, even though it's not part of the mathquill config object and uses replacements. I also didn't want to create another new plugin just for this small change.

### DIFF
--- a/localization/en.ftl
+++ b/localization/en.ftl
@@ -325,6 +325,8 @@ custom-mathquill-config-opt-extendedGreek-name = More Greek Letters
 custom-mathquill-config-opt-extendedGreek-desc = Enables replacements for all supported greek letters
 custom-mathquill-config-opt-lessFSpacing-name = Less Spacing Around "f"
 custom-mathquill-config-opt-lessFSpacing-desc = Reduces extra spacing around the letter "f"
+custom-mathquill-config-opt-strongMismatchedBrackets-name = strong mismatch brackets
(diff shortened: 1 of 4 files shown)


ronwnor:
It's effectively my first time using git/ multiplayer github so sorry if I'm doing something stupid

I'm also bad at naming maybe this should be named "disable brackets matching with parens" or something idk ><

jared-hughes:
Seems like brackets and parens should _never_ match with each other in Desmos since it always gives an error.
Blame goes back to https://github.com/mathquill/mathquill/pull/391.

Would you be willing to PR this change to https://github.com/desmosinc/mathquill? I could help get it merged in.

For backwards compatibility, I suggest taking either of the following two routes:
1. Add an option `restrictAllMismatchedBrackets` to the config, that enables `restrictMismatchedBrackets` and disables the part about keeping `(]` and `[)` in.
2. Extend the option `restrictMismatchedBrackets` to take a third option like `"none"`. (I prefer this since it removes the question of what happens if `restrictMismatchedBrackets` is specified false but `restrictAllMismatchedBrackets` is true)

The configuration option is declared in `publicapi.ts`, and applied in `commands.ts`. A test would go after the suite `restrictMismatchedBrackets` in `typing.test.js` (you can run the tests with `make test`)

ronwnor:
thanks feny! <3

PR #865, opened 2024-04-23
Remove home link disable on panic popover

SlimRunner:
I presume this change was made a while ago when the panic button was on top of Desmos' home link. I was checking the git history of this line, and it seems the last change was about a year ago in May, so I think it lines up. Either way the purpose of this change is because currently the link is unclickable until the popover is open, but then it is blocked, so it renders the home link unusable.

### DIFF
--- a/src/panic/panic.html
+++ b/src/panic/panic.html
@@ -92,10 +92,6 @@
   body.dsm-panic-open #dsm-panic-reopen-button {
     display: none;
   }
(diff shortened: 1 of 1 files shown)


jared-hughes:
The benefit to being to lazy to fix the replacement error is that you figure out problems with the panic handling :).

Pretty sure the integration test errors are just because shift-enter-newline is broken, so I'll just override and merge this in now.

PR #866, opened 2024-04-25
Add allowUndo to refreshState for performance-view

FenyLabs:
preserves undo history

### DIFF
--- a/src/plugins/performance-info/index.ts
+++ b/src/plugins/performance-info/index.ts
@@ -45,7 +45,7 @@ export default class PerformanceInfo extends PluginController {
 
   refreshState() {
     this.cc._showToast({ message: "Refreshing graph..." });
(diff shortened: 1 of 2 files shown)


FenyLabs:
integration test fails because of hide-errors, fixed in #867

src/plugins/performance-info/index.ts
```diff
@@ -45,7 +45,10 @@ export default class PerformanceInfo extends PluginController {
 
   refreshState() {
     this.cc._showToast({ message: "Refreshing graph..." });
-    this.calc.setState(this.calc.getState());
+    // should this be using killWorker instead?
```
jared-hughes, line 48: If it works it works ¯\_(ツ)_/¯

jared-hughes (approved):
Nice

PR #867, opened 2024-04-25
Fix replacement in hide-errors

FenyLabs:


jared-hughes:
Thanks! Very helpful.

Looks like this comes from a change where Desmos won't automatically add sliders anymore if you press "enter" with more than four slider variable suggestions.

jared-hughes (approved):

FenyLabs:
> Thanks! Very helpful.
> 
> Looks like this comes from a change where Desmos won't automatically add sliders anymore if you press "enter" with more than four slider variable suggestions.

Presumably to avoid spamming variables when you mistype something like "folder"

Issue #898, opened 2024-07-01
Hide-errors again breaking tooltips


jared-hughes:
Clicking doesn't show sticky tooltips, which is a big problem on mobile where you can't hover.

PR #899, opened 2024-07-19
Future-proof BEV replacements, and fix tooltips with hide-errors

jared-hughes:


jared-hughes:
Oops accidentally pushed an extra commit. Whatever. Fixes https://github.com/DesModder/DesModder/issues/898.

PR #902, opened 2024-07-20
Shape Generator Plugin

lafkpages:
Adds a simple utility plugin that helps quickly generating shapes and their equations in Desmos. Currently only supports rotatable ellipses and rectangles, but I'll be adding more soon.

The idea for this plugin came to me while trying to do some art in Desmos. I needed to add some shapes on top of an image I had for reference. This plugin would save me from having to write the equations for different shapes out by hand, substitute variables to easily tweak them, and then substitute in the actual values once I'm done.


package.json
```diff
@@ -81,12 +81,14 @@
...
     "@codemirror/state": "^6.4.1",
     "@codemirror/view": "^6.26.0",
+    "@cortex-js/compute-engine": "^0.25.1",
     "@ffmpeg/core": "^0.11.0",
     "@ffmpeg/ffmpeg": "^0.11.6",
     "@fluent/bundle": "^0.18.0",
     "@lezer/common": "^1.2.1",
     "@lezer/lr": "^1.4.0",
     "client-zip": "^2.4.4",
+    "compute-engine-to-desmos-latex": "^0.0.1",
```
jared-hughes, line 91: It's better to bring this package directly into the repository (which is a monorepo). Otherwise, development is a dance (`npm link` and stuff is tricky), and updates to the package would require many more steps (make the package change, release it to npm, update the version in desmodder, etc.)

If desired, we could still publish it (and it could even be in the `@desmodder` namespace), but I suspect there's not much demand for it.

jared-hughes:
This is cool! Reminds me of [Desmos Lock](https://www.youtube.com/watch?v=3Yda5CoTSJA), but this has a much better UX.

It's such a seamless process of integration with Desmos. The "OK" button has such a nice placement. The insertion of extra sliders and the main expresion (and their cleanup) works well, though I haven't tested it thoroughly. 

Without looking too deeply at the code, I see this PR as consisting of three parts:
1. Adding buttons on the "Add item" menu to insert several pre-populated expressions at once
   - To support part 2, this marks one expression as "OK" and the rest as deleteable.
2. Adding an "OK" button to certain expressions that, when pressed:
    - replaces slider variables with their definitions in the "OK" expression
    - deletes the marked deleteable dependencies
    - removes the "OK" button
3. Simplifying the "OK" expression (hopefully this is all that CortexJS is used for)

I'd be interested in merging this _without_ part 3 (and likely merging it separately afterwards). In my experience, a big PR that does several separate things ends up neglecting some part of some of them. 

I'd rather focus on part 1 and 2 for this PR, taking care to deal with edge cases like hidden folders and variables with duplicate names (defined as functions, sliders, tables, regressions, etc). Then we can later focus on part 3, taking care to deal with all the involved edge cases.

I'm also inserted (as a separate PR) of allowing users to define their own "shapes" (which only affects part 1). A custom shape could be stored as a graph like https://www.desmos.com/calculator/dyymzrjd8i. This is pretty similar to https://github.com/DesModder/DesModder/pull/748, so I'll see what I could do to polish that up into a working version (unless @radian628 wants to wrap it up) so we could re-use a bunch of the machinery.

package.json
lafkpages: I'm using it in several other projects of mine, that's why I was publishing it. In fact, I nearly put it in a monorepo for another project. But if it can be put in here and still be published on `@desmodder`, that'd be fine.

lafkpages:
> I'd be interested in merging this _without_ part 3 (and likely merging it separately afterwards). In my experience, a big PR that does several separate things ends up neglecting some part of some of them.

Alright, I've removed all CortexJS logic for now.

> I'm also interested (as a separate PR) of allowing users to define their own "shapes" (which only affects part 1). A custom shape could be stored as a graph like https://www.desmos.com/calculator/dyymzrjd8i.

Yeah that could be interesting.

bunfig.toml
jared-hughes, line 1: I'd rather not have extraneous files checked in. Can you remove this?

jared-hughes: I'm not sure how I feel about introducing bun considering it's a huge dependency. But I'd likely approve a PR entirely replacing jest with bun (jest has been a pain to configure).

localization/en.ftl
```diff
@@ -19,7 +19,7 @@ category-integrations-name = Integrations
 
 ## GLesmos
 GLesmos-name = GLesmos
-GLesmos-desc = Render implicits on the GPU. Disabled on tab reload. Can cause the UI to slow down or freeze in rare cases; reload the page if you have issues.
+GLesmos-desc = Render implicits on the GPU. Disabled on tab reload. Can cause the UI to slow down or freeze in rare cases ; reload the page if you have issues.
```
jared-hughes, line 22: Extra space a mistake?

localization/en.ftl
```diff
@@ -282,10 +282,10 @@ multiline-opt-multilinifyDelayAfterEdit-name = Edit Delay (ms)
 multiline-opt-multilinifyDelayAfterEdit-desc = Multiline expressions should be updated after no edits are made for this number of milliseconds.
 multiline-opt-spacesToNewlines-name = Spaces to Newlines
 multiline-opt-spacesToNewlines-desc = Convert groups of 3 spaces into newlines. These can be automatically created with Shift+Enter.
-multiline-opt-determineLineBreaksAutomatically-name = Auto Insert Linebreaks 
+multiline-opt-determineLineBreaksAutomatically-name = Auto Insert Linebreaks
```
jared-hughes, line 285: Do you have something automatically removing trailing spaces? That could be a separate PR to keep this PR smaller.

localization/es.ftl
```diff
@@ -360,3 +360,7 @@ better-navigation-opt-ctrlArrow-name = Atajo Ctrl+Flecha.
 better-navigation-opt-ctrlArrow-desc = Te permite utilizar Ctrl+Flecha y Ctrl+Shift+Flecha para avanzar bloques de texto más rápidamente. De manera similar puedes usar Ctrl+Retroceso para borrar bloques.
 better-navigation-opt-scrollableExpressions-name = Expresiones Desplazables
 better-navigation-opt-scrollableExpressions-desc = Añade una barra de desplazamiento horizontal. Mayormente destinado para facilitar la navegación en móvil.
+
+## Shape Generator
+shape-generator-name = Generador de Formas
+shape-generator-desc = Facilita la creación de equaciones para distinas formas.
```
jared-hughes, line 366: Missing the show-sliders name and desc

src/plugins/shape-generator/index.ts
```diff
@@ -0,0 +1,407 @@
...
+          latex: "r_{xEllipseGenerator}",
+        });
+        const ryHelper = this.calc.HelperExpression({
+          latex: "r_{yEllipseGenerator}",
+        });
+        const angleHelper = this.calc.HelperExpression({
+          latex: "A_{ellipseGenerator}",
+        });
+
+        angleHelper.observe("numericValue", () => {
```
jared-hughes, line 60: This seems to assume that `angleHelper` gets its `observe` triggered only after all helpers have their numeric value resolved. While that seems to be correct, a comment to this effect could help future debugging.

src/plugins/shape-generator/index.ts
```diff
@@ -0,0 +1,407 @@
...
+          const y = yHelper.numericValue;
+          const rx = rxHelper.numericValue;
+          const ry = ryHelper.numericValue;
+          const angle = angleHelper.numericValue;
+
+          if (isNaN(x) || isNaN(y) || isNaN(rx) || isNaN(ry) || isNaN(angle)) {
+            return;
+          }
+
+          // TODO: HelperExpression is missing unobserve types
```
jared-hughes, line 71: pending TODO (another one below)

src/plugins/shape-generator/index.ts
```diff
@@ -0,0 +1,407 @@
...
+    },
+  ];
+
+  private expressionsMutationObserver: MutationObserver | null = null;
+
+  afterEnable() {
+    // Construct stylesheet
+    const style = document.createElement("style");
+    style.textContent = `
+      /* Disable the save button while editing a shape */
```
jared-hughes, line 162: These styles should be moved to a .less file. I think you can just create `shape-generator.less`,  put the code in there, and `import './shape-generator.less` from this index.ts file.

If these styles shouldn't run when the shape-generator plugin is disabled because they break other functionality, then they shouldn't run when the shape-generator plugin is enabled either.

If they merely don't _need_ to run when the plugin is disabled, there's not a concern with moving it to a .less file since a few more styles doesn't really hurt performance.

src/plugins/shape-generator/index.ts
```diff
@@ -0,0 +1,407 @@
...
+
+      .shape-generator-ok-btn {
+        margin-top: 36px;
+        margin-left: 1px;
+      }
+    `;
+    style.id = "shape-generator-styles";
+    document.head.appendChild(style);
+
+    this._onDispatchHandler = this.calc.controller.dispatcher.register(
```
jared-hughes, line 192: Ah sorry I should've pinged you about this. In https://github.com/DesModder/DesModder/pull/909/files I added a way to make custom dispatch handling easier (hopefully) to work with. Using that API would in particular make the undo/redo behavior correct by default.

src/plugins/shape-generator/index.ts
```diff
@@ -0,0 +1,407 @@
...
+      this.onDispatch.bind(this)
+    );
+
+    // Listen for expression to add the OK button to
+    this.expressionsMutationObserver = new MutationObserver((mutations) => {
```
jared-hughes, line 197: I don't personally like MutationObserver stuff because they make HTML the source of truth, and we can usually hook into the JS. But if it works it works.

src/plugins/shape-generator/index.ts
```diff
@@ -0,0 +1,407 @@
...
+  }
+
+  get isEditingShape() {
+    return this._isEditingShape;
+  }
+
+  set isEditingShape(value) {
+    this._isEditingShape = value;
+
+    document.body.dataset.shapeGeneratorIsEditingShape = value.toString();
```
jared-hughes, line 304: Since this is just a boolean, a CSS class feels more appropriate. We have precedent with `"dsm-better-nav-scrollable-expressions"`. But I doubt it's worth changing it. Both implementation approaches behave about the same.

src/plugins/shape-generator/latex/ellipse.ts
```diff
@@ -0,0 +1,121 @@
...
+    id: "shape-generator-ellipse-angle-point",
+    type: "expression",
+    latex:
+      "\\left(f_{ellipseGeneratorAngleX}\\left(A_{ellipseGenerator}\\right),f_{ellipseGeneratorAngleY}\\left(A_{ellipseGenerator}\\right)\\right)",
+    dragMode: "X",
+    color: "",
+  },
+] satisfies Desmos.ExpressionState[];
+
+// Ensure that the expressions that will be hidden are at the bottom of the list
```
jared-hughes, line 121: This looks like the comment is placed on the end of the file, instead of modifying what's before it. Could you move it to before the first `hidden` expr, or maybe before the whole list?

src/plugins/shape-generator/index.ts
lafkpages: Oh, but that's an issue of `HelperExpression`s missing the type annotations for the `unobserve` function. I'm not sure what the best way to type this would be, to simply copy-paste the `unobserve` types from `Calc`? Or to make a shared type like `interface Observable` that the rest of types extend?

bunfig.toml
lafkpages: Yeah sorry I'll remove this. I had added it because I was using Bun as my package manager, since it can automatically install from an existing `package-lock.json` and is way faster than npm.

Also regarding Bun being a huge dependency, imo it's worth it, since Bun on its own is basically a better replacement for Node, Jest, npm, and esbuild.

localization/en.ftl
lafkpages: Yeah, didn't mean to touch other translations sorry. Will undo.

bunfig.toml
jared-hughes: I hear that. It sounds great. But I'm not personally planning to commit X hours to switching the codebase over to it.

src/plugins/shape-generator/index.ts
jared-hughes: I'd typically use .replacements for this, but those are hard to learn and perhaps a little more fragile than MutationObserver (since MutationObserver would break if the HTML output changes, but the .replacements would break if either the HTML output changes or if the underlying DCGView implementation changes).

src/plugins/shape-generator/index.ts
jared-hughes: Oh, that's tricky since the types are in the `desmos` package (which is 4 years old, incorrect in parts, and NOT published by Desmos). I'd probably just toss in an `as any` and a comment explaining that the `desmos` package is wrong here.

src/plugins/shape-generator/index.ts
```diff
@@ -321,9 +314,12 @@ export default class ShapeGenerator extends PluginController<{
     }
   }
 
-  onDispatch(e: DispatchedEvent) {
+  handleDispatchedAction(e: DispatchedEvent) {
     if (e.type === "toggle-add-expression") {
-      this.addExpressionMenuHandler();
+      // Wait for addExpressionBtn.ariaExpanded to be updated
+      queueMicrotask(() => {
```
jared-hughes, line 327: I haven't heard of `queueMicrotask` before. Neat!

It's not clear to me exactly when that JS callback is ran. Presumably after the dispatched action. But if one function call dispatched two actions in a single task, I don't know if it would run after both actions. There's the same problem with the `setTimeout(() => {}, 0)` speckled throughout the code. A more framework-brained way to do this is replacing `queueMicrotask(() => {})`` with `this.cc.runAfterDispatch(() => {})`, but I don't think you need to make this change.

src/plugins/shape-generator/index.less
```diff
@@ -20,7 +18,7 @@
 }
 
 /* Hide sliders if the setting is disabled */
-[data-shape-generator-show-sliders="false"]
+.dsm-shape-generator-hide-sliders
```
jared-hughes, line 21: Ah, this shows one downside (upside?) to classes. With datasets, it's easy to get a tristate of (false, true, missing). But it looks like you figured it out.

src/plugins/shape-generator/index.ts
lafkpages: Ah, didn't know `runAfterDispatch` was a thing. I'll try it out and see if it works, it seems like exactly what I wanted and it also makes the code more readable.

src/plugins/shape-generator/index.ts
lafkpages: Alright, I guess I'll type assert for now. But maybe a PR could be made to `@types/desmos` fixing these types? Or just fork the package, and add it to the DesModder monorepo, and fix them there.

src/plugins/shape-generator/index.ts
lafkpages: Wait, is it `this.cc.runAfterDispatch`? It doesn't seem to be typed...

src/plugins/shape-generator/index.less
lafkpages: Yup, since the styles only need to be applied to one of those states, a class also works just fine here.

src/plugins/shape-generator/index.ts
lafkpages: Ok in DevTools it seems to be `Calc.controller.runAfterDispatch`, and I tried that in my plugin and it seems to work. But yeah, the types are missing.

src/plugins/shape-generator/index.ts
lafkpages: So do you think I should use replacements here? MutationObserver seems to work just fine, and I have no idea how these replacements would work, although I could have a look at other plugins if I had to use them.

bunfig.toml
lafkpages: If you want I could make a PR? Or several smaller PRs? Switching the tests to Bun could also improve CI workflow usage times.

But anyway, I'll resolve this as it's not really related to this plugin anymore.

jared-hughes:
The `display: none !important;` is problematic when the expressions list can be scrolled. Desmos does measurements and calculations on the vertical height to avoid showing items that are off-screen. The `display: none` breaks that.

My first thought for a fix was marking the secret helpers as `secret: true`, which hides them in the expressions list (while `hidden: true` hides them on the graphpaper). But there's a slight problem that anyone with ?authorFeatures enabled (such as through the Calculator Settings plugin) would be able to see the secret helpers, and it would clutter their vertical space.

A simple approach would be to use a collapsed secret folder (at the end of the expressions list) to hide all the expressions that you don't want to show to the user. That way, it's hidden from most users, and those with ?authorFeatures would only have a slight vertical affliction. (This may also help cleaning up -- the OK button could just delete the folder in a method analogous to hitting the "X").

A complicated approach would be using some other approach to hide the expressions. One way would be to set their item models to have `filteredBySearch = true`, such as by `Calc.controller.getItemModelByIndex(0).filteredBySearch = true`. That  would break on a Desmos Ctrl+F search (which mutates the `filteredBySearch` booleans), but I imagine there's some workaround there.

lafkpages:
> My first thought for a fix was marking the secret helpers as `secret: true`, which hides them in the expressions list

I thought I tried that but I could only get folders to be hidden and not individual expressions?

> But there's a slight problem that anyone with ?authorFeatures enabled (such as through the Calculator Settings plugin) would be able to see the secret helpers, and it would clutter their vertical space.

True

> A simple approach would be to use a collapsed secret folder (at the end of the expressions list) to hide all the expressions that you don't want to show to the user.

Yeah, I tried that before. Only problem is that the only way to add my expressions into a folder is by using `Calc.setState()`. Specifying `folderId` in `setExpression` does quite literally nothing. But having to `getState()`, filter, update, and `setState()` is painful.

jared-hughes:
I pushed a commit to remove the disabling of the "save" button because there's at least 3 other ways to save a graph.

A few final bugs:
1. On an [empty graph with ?showIDs](https://www.desmos.com/calculator/qycm78tiwi?showIDs), the expression list HTML is more blank, and the plugin does not load properly (`Uncaught Error: Could not find the expression list at _ShapeGenerator.afterEnable  (index.ts:232:13)`)
2. Clearing the graph (cog menu > Delete All) or opening a new graph from the sidebar (which does a `Calc.getState(...)`) does not reset `isEditingShape`, so the "ellipse" and "rectangle" buttons remain grayed out and can't be used.
3. New expressions always get added at the end, instead of after the selected item. (Don't worry about fixing this one. I can do it later).

lafkpages:
> 1. On an [empty graph with ?showIDs](https://www.desmos.com/calculator/qycm78tiwi?showIDs), the expression list HTML is more blank, and the plugin does not load properly (`Uncaught Error: Could not find the expression list at _ShapeGenerator.afterEnable  (index.ts:232:13)`)

Oops, that should be an easy fix.

> 2\. Clearing the graph (cog menu > Delete All) or opening a new graph from the sidebar (which does a `Calc.getState(...)`) does not reset `isEditingShape`, so the "ellipse" and "rectangle" buttons remain grayed out and can't be used.

Is there a dispatch for clearing, resetting, or opening a new graph that I can hook into?

> 3\. New expressions always get added at the end, instead of after the selected item. (Don't worry about fixing this one. I can do it later).

Alright

lafkpages:
@jared-hughes are you still going to fix this? Or should I? And what about the dispatches?

SlimRunner:
> Is there a dispatch for clearing, resetting, or opening a new graph that I can hook into?

@lafkpages Clearing, resetting, and opening a new graph always dispatch the `set-state` event which comes with the new `state: GraphState` as payload. The state is `undefined` when clearing and has the new graph state otherwise.

You can find what these dispatchers are by calling
```js
Calc.controller.dispatcher.register(e => console.log(e));
```
on an empty graph. This and a lot more useful info on dispatchers is available in [dispatch-handlers.md](https://github.com/SlimRunner/DesModder/blob/5cdec1a44b6975341444f015dcbb2b2376e27673/docs/dev/dispatch-handlers.md)

localization/es.ftl
```diff
@@ -360,3 +360,9 @@ better-navigation-opt-ctrlArrow-name = Atajo Ctrl+Flecha.
...
```
SlimRunner, line 366: There are two little typos here. I think you meant
```suggestion
shape-generator-desc = Facilita la creación de ecuaciones para distintas formas.
```

SlimRunner:
@lafkpages I also looked into the dispatchers. Using [DesThree](https://github.com/jared-hughes/DesThree) and some debugging in combination with DesModder dispatch guide I came up with the following:
```js
(function() {
  'use strict';
  const expr = Calc.controller.createItemModel({
    latex: 'x^3',
    type: 'expression',
    id: Calc.controller.generateId(),
    color: Calc.controller.getNextColor()
  });
  Calc.controller._toplevelNewItemAtSelection(expr, {shouldFocus: true});
  Calc.controller._closeAddExpression();
  Calc.controller.dispatch({});
}());
```
Obviously this is no good for this repo, so I looked further and implemented my own "new polynomial" dispatcher that adds a random polynomial at the selected expression in [my personal sample plugin branch](https://github.com/SlimRunner/DesModder/tree/example/simple-plugin). Feel free to [inspect the diff](https://github.com/SlimRunner/DesModder/commit/04984d8e81eedfe098b7096ac62b114f2c93d30b) if you want to adapt it for your plugin.

localization/es.ftl
lafkpages: I'm literally Spanish how could this happen 🫠

SlimRunner: Don't worry. Spanish is also my mother tongue, and sometimes I write "the" instead of "de" 💀

PR #911, opened 2024-07-27
Fix/minor gui issue

### LINKED ISSUE #910: Version should be relative to inside of scrolling container
image

SlimRunner:
Fixes #910 

I added a grid display property for the container of the title and version. I left aligned the version div with margins within its cell.

I also added the missing entry for "Fast screenshot" to the Spanish translation.


jared-hughes (approved):
Thanks! Looks great

PR #923, opened 2024-08-15
Fix/style glitch

SlimRunner:
This patch fixes
- missing style from blue buttons
- missing text from all buttons
- intellisense showing an empty list

Notes:
Let me know if the way I fixed the change from `this.children` to `this.prop.children` is OK. I could not figure out how the children property was assigned formerly (I assumed through Desmos code from stepping through the code), so I just took the shotcut and added children as a component property of the appropriate components as type `any`.

I have tested only on Chrome. I will test later on Firefox.

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -7,6 +7,7 @@ export default class Button extends Component<{
   class?: MaybeClassDict;
   onTap: (e: Event) => void;
   disabled?: boolean;
+  children?: any;
 }> {
   template() {
     return (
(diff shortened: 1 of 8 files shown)


SlimRunner:
From a quick test on Firefox seems everything is fine there too.

SlimRunner:
Oh one more thing I forgot to mention. When fixing the classes, I tried to regenerate the file _overrides.less because it also contained some `dcg-btn-primary`, but I could not so against the warning there about not editing manually I just edited manually.

jared-hughes:
Manual edit is fine in this case, though I'll likely run an automated generation afterwards to refresh the rest of the styles.

jared-hughes (approved):
Looks good! Thanks for doing this.

PR #924, opened 2024-08-16
Prepare v0.13.10

jared-hughes:
@SlimRunner I added some extra instructions in clean-css.mjs here with regards to generating the css overrides. Downloading the css isn't yet automated in the same way that `setup-unit.js` automatically downloads the JS.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,22 @@
 # DesModder Changelog
 
+## 0.13.10
+
+Breaking Changes:
+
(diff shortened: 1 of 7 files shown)


SlimRunner:
Ah, good thing we have the feed in Discord. Should I make a branch to add this?
```
builtin-settings-opt-showIDs-name = Mostrar IDs
# Unchanged
builtin-settings-opt-showIDs-desc = {""}
```
I goes below line 139 in `es.ftl`.

Or you could do it since it is so small.

jared-hughes:
Ack, I just submitted to Chrome. It'll have to go in the next release. I can take care of it.

SlimRunner:
No problem. I will do push the PR later tomorrow.

PR #925, opened 2024-08-18
Git ignore .DS_Store files on macOS

lafkpages:
Yes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@ dist
 dist-ts
 *.zip
 coverage
+
+# macOS Finder metadata files
+.DS_Store


FenyLabs:
I would recommend adding .DS_STORE to your own personal `core.excludesFile`, which is probably in `~/.config/git/ignore` but I'm not super sure for mac. .DS_STORE files are specific to your operating system, not the repo, so generally I wouldn't commit them to the source-controlled .gitignore.

jared-hughes:
I don't mind the extra .gitignore in this case. It doesn't do much harm. The .DS_STORE is purely metadata.

jared-hughes (approved):

PR #926, opened 2024-08-19
Fix draw order lookup in Intellisense

SlimRunner:
This patch fixes the Intellisense lookup which currently fails to sort the items by drawing order because Desmos removed a property from Calc.

The property removed was the `Calc.controller.listModel.drawOrder`. It's changed to `Calc.controller.listModel.drawLayers[...].drawOrder`. I am not sure why `drawLayers` is an array instead of an object, but I opened several graphs and they all had only one element. Maybe it's future proofing for later, but whatever the case I hardcoded the zero in the intellisense plugin.

### DIFF
--- a/src/globals/Calc.ts
+++ b/src/globals/Calc.ts
@@ -269,7 +269,7 @@ interface CalcPrivate {
       }[];
       __itemIdToModel: Record<string, ItemModel>;
 
-      drawOrder: string[];
+      drawLayers: { drawOrder: string[] }[];
     };
     _addItemToEndFromAPI: (item: ItemModel) => void;
     _showToast: (toast: Toast) => void;
--- a/src/globals/Calc.ts
+++ b/src/globals/Calc.ts
@@ -269,7 +269,7 @@ interface CalcPrivate {
       }[];
       __itemIdToModel: Record<string, ItemModel>;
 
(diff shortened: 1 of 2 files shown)


jared-hughes:
I suspect it's for the "in front of everything" checkbox: https://www.desmos.com/calculator/ow9zkgenlp

SlimRunner:
Oh shut. I could cause the ordering to be unstable. I think a simple reduction should suffice.

SlimRunner:
I think that should be it. If you need another change let me know.

jared-hughes:
Now I'm wondering why Intellisense uses the draw order at all. Shouldn't it just use index order in the expressions list (@radian628?)?. Now that they're distinct, this actually matters. That would simplify this sorting code.

SlimRunner:
I was so focused on leaving the code as unaltered, I didn't stop to think. Yeah, now (more than before) would make more sense to sort by index.

SlimRunner:
Oh, here is a think I did not consider. @radian628 , the intellisense menu gets updated each time it is invoked (by typing), right?

I ask because if it doesn't then I just introduced indexes that can go stale.

jared-hughes (approved):
So simple now! Looks great

PR #927, opened 2024-08-19
Add option to hide folder toggles

SlimRunner:
The changes allow to hide the folder toggles added by "advanced styling" and "author features".

I read [a conversation in the Desmos Discord server](https://discord.com/channels/655972529030037504/899774334602719272/1274770255444709407) about the toggles being slightly cumbersome and I agree and have been misclicking them for a while now. So, this change is to address that.

I used the code already in present in Compact View, and I think it fits well there.


SlimRunner:
By the way for those who prefer the non-compact view, but want to hide the folders you can simply slide the "remove spacing" slider all the way left.

Also, I made it default true because those two toggles do not play nice with the compact view anyway. They clip through the expression.

src/plugins/compact-view/compact.less
```diff
@@ -211,3 +211,11 @@
     }
   }
 }
+
+.compact-view-enabled.hide-folder-toggles {
+  .dcg-action-toggle-secret-folder,
+  .dcg-action-toggle-in-front-of-everything {
+    display: none;
+    visibility: hidden;
```
jared-hughes, line 219: Is `visibility: hidden` needed? Should be redundant with `display: none` afaict.

jared-hughes (changes requested):
Nice! Useful and simple improvement.

I wonder about hiding the checkbox when "advanced styling" and "author features" are both off. Not sure if that's even a good change, and if it was it might not be worth the effort.

src/plugins/compact-view/compact.less
SlimRunner: I wasn't sure if it was needed. I can remove it.

SlimRunner: It was redundant. I just removed it

SlimRunner:
> Nice! Useful and simple improvement.
> 
> I wonder about hiding the checkbox when "advanced styling" and "author features" are both off. Not sure if that's even a good change, and if it was it might not be worth the effort.

How about graying it out? Kinda like GLesmos in 3D?

SlimRunner:
That should do it. It shows only when either of the related options is true.

jared-hughes (approved):
Thanks. I added two more commits to get the actual value of `Calc.advancedStyling`.

PR #928, opened 2024-08-29
Fix broken replacements

SlimRunner:
It seems like an oddly simple fix, but I am not 100% confident that was it (on the off change that I searched the wrong new replacement). The error was correctly identified by the panic system.
> "Show style circles and expression footers for Text Mode statements"

The page crashes if you try to open text mode when there is an expression with a style circle present.

### DIFF
--- a/src/components/desmosComponents.ts
+++ b/src/components/desmosComponents.ts
@@ -171,6 +171,7 @@ export abstract class ExpressionViewComponent extends ClassComponent<
   }
 > {}
 
+// `?` to avoid a crash if the replacement fails
 const ExpressionView = window.DesModderFragile?.ExpressionView;
 
 export abstract class IconViewComponent extends ClassComponent<{
@@ -178,6 +179,7 @@ export abstract class IconViewComponent extends ClassComponent<{
   controller: CalcController;
 }> {}
 
+// `?` to avoid a crash if the replacement fails
 export const ImageIconView = window.DesModderFragile?.ImageIconView;
 
 interface ModelAndController {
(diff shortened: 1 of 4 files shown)


SlimRunner:
Do not mind the force push. I rushed to push what I thought was a fix for the newest error (as of today I believe), and it wasn't fixed so I just recalled my commit.

SlimRunner:
It seems `this.model` within the `shouldShowSliderPrompt` function no longer works. I presume it is nested differently now, but whatever the case now the function receives a parameter which stores the current expression. I can think of ways it could break, but please take a careful look at that commit.

SlimRunner:
The last commit fixes #929.

SlimRunner:
For some reason the previous comment did not auto-linked the issue to be closed by this PR.

SlimRunner:
I force-pushed the same commit. I simply forgot to add the `inside` keyword in the replacement. Although, this little mistake made me realize it isn't really needed, is it? Either way I updated the commit for consistency.

src/components/desmosComponents.ts
```diff
@@ -171,14 +171,14 @@ export abstract class ExpressionViewComponent extends ClassComponent<
   }
 > {}
 
-const ExpressionView = window.DesModderFragile.ExpressionView;
+const ExpressionView = window.DesModderFragile?.ExpressionView;
```
jared-hughes, line 174: Why a `?` here? Just to avoid a crash when the replacement fails?

jared-hughes (approved):

src/components/desmosComponents.ts
jared-hughes: Nevermind, I can read. It's to fix #929.

jared-hughes:
Yep the `inside` fluff doesn't do anything. The parser just looks for the next `inline code` after the `*Find*` token.

PR #930, opened 2024-09-17
Improve Japanese localization

64qq:
Because the original ja.ftl file contained some unnatural expressions that were machine-translated and lacked some localization, I have made overall improvements, including correction of mistranslations and unification of inconsistent notations.


jared-hughes:
ha! The in-browser translate makes the diff funny.
image

jared-hughes:
Thanks for the improvements!

Test failure seems unrelated, likely fixed by #928, so I'm merging.

PR #931, opened 2024-09-18
Fix wrong screenshot scaling

### LINKED ISSUE #920: Video Creator catpures do not scale down below "native resolution"
I was doing some testing in #919 and found a bug in the image capture. I tried searching for it because I think I had heard someone saying this, but I could not find it.

Anyway, like the title says if the target size is modified in such a way that if it results in a number smaller than the native resolution, DesModder preserves the ratio of the target resolution but uses the width of the native resolution.

For example if I have a graph with a viewport resolution of 1420x909 and set the target resolution to 480x270, then the resulting capture with the "target same pixel ratio" unchecked is 1420x798. This has approximately the correct aspect ratio of 16:9.

SlimRunner:
This PR fixes #920 with some caveats (see below)

Changes made
- screenshot mode set to "contain"
- pixel ratio is clamped to 1
- stop preview play when "delete all" is clicked

I believe there is no point in allowing pixel ratios that are less than 1 now that Desmos doesn't seem to support them anymore. As pointed out in the issue linked, this causes some screenshots below native resolution to actually increase in resolution. The drawback—which I think is reasonable—is that now these screenshots are no longer matched with what you see on screen (as if the same pixel ratio was checked on) as opposed to the default behavior.

The "contain" option fixes alignment problems with screenshots taken without fast screenshots. For some reason, these were preserving x instead which is odd because the default behavior according to the API is "contain".

### DIFF
--- a/src/plugins/video-creator/backend/capture.ts
+++ b/src/plugins/video-creator/backend/capture.ts
@@ -19,13 +19,13 @@ async function captureFrame(
   const width = vc.getCaptureWidthNumber();
   const height = vc.getCaptureHeightNumber();
   const targetPixelRatio = vc.getTargetPixelRatio();
-  // resolves the screenshot as a data URI
   const size = {
     width: width / targetPixelRatio,
     targetPixelRatio,
     height: height / targetPixelRatio,
     preserveAxisNumbers: true,
   };
+  // resolves the screenshot as a data URI
   const screenshot = vc.cc.is3dProduct() ? screenshot3d : screenshot2d;
   return await Promise.race([screenshot(vc, size), vc.awaitCancel()]);
 }
@@ -57,6 +57,7 @@ async function screenshot2d(vc: VideoCreator, size: ScreenshotOpts) {
     ...size,
     showLabels: true,
     mathBounds: clampedMathBounds,
+    mode: "contain" as "contain",
   };
   if (vc.fastScreenshots) {
     return await new Promise<string>((resolve) => {
--- a/src/plugins/video-creator/backend/capture.ts
+++ b/src/plugins/video-creator/backend/capture.ts
@@ -57,7 +57,7 @@ async function screenshot2d(vc: VideoCreator, size: ScreenshotOpts) {
     ...size,
     showLabels: true,
     mathBounds: clampedMathBounds,
-    mode: "contain" as "contain",
+    mode: "contain" as const,
   };
   if (vc.fastScreenshots) {
     return await new Promise<string>((resolve) => {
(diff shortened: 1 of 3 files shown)


SlimRunner:
I discussed it in the related issue, but let me know if [this integration test (this is the diff)](https://github.com/SlimRunner/DesModder/compare/56c7dcb7a24c2cda5db70623200609b4dfac50f7...43981318f5d2b3fa1c38a6c170dbc64455cb70f3) is worth adding to this PR.

The reason I am asking and not just pushing the change is that the newer test is not as _clean_ as the original. At a glance my concerns are:
- the `Menuing` test no longer exits cleanly
- the new `Resizing screenshot` test takes roughly x1.5 to x2 longer than `Menuing`
  - increased from ~2s to ~6s in Ubuntu and from ~5s to ~10s in Windows.
- it tests only for the default puppeteer resolution of 480x557
- if it fails, it can be annoying to appease (because it has multiple points of failure), but that also means it is thorough.

jared-hughes:
Will adding the integration test save more time in the future than it will cost? I'm guessing not (since it sounds like there's some things you need to do to stop it from flaking), so I'd lean towards not adding it.

jared-hughes:
I had trouble understanding what was going on, but I think I've got it now. Both before and after the PR, specifying 480x270 on a 1420x909 window would give you a 1420x798 image output. This was previously implicit by whatever Desmos internals, and now it's explicit by clamping to 1. Before the PR, the image wouldn't match what's on the screen --- too many grid lines, for example. After the PR, you get the same result as if you specified 1420x798 in the first place.

src/plugins/video-creator/backend/capture.ts
```diff
@@ -57,6 +57,7 @@ async function screenshot2d(vc: VideoCreator, size: ScreenshotOpts) {
     ...size,
     showLabels: true,
     mathBounds: clampedMathBounds,
+    mode: "contain" as "contain",
```
jared-hughes, line 60: `as const` also works here.

jared-hughes (approved):

src/plugins/video-creator/backend/capture.ts
SlimRunner: Nice! That is cleaner. I will change it.

SlimRunner:
I was gonna squash, but it would mess up the flow of the dialog here. So I leave it up to you when you merge.

About the error. I also had some issues understanding what was going on, but, yeah, (1) framing of non-fast screenshot was messed up, (2) and, as you said, the scaling down below native resolution sometimes did not work because a pixel ratio of < 1 did nothing (anymore).

PR #932, opened 2024-09-18
Fix compact-view default font ratio

SlimRunner:
This is a quality of life change.

The default font size for expressions is 110% above the regular font size. That means that evaluation containers have a font size of 16 px while expressions have a font size of roughly 17.6 px. Default math font size now reflects the actual default (16 px).

Now when the spacing slider is all the way to the left, Desmos has a _default feel_. Previously you had to choose between the expressions or evaluation containers having a larger font-size than they normally do. This change was made to remove visual snags if all you want to enable is the "hide folder toggle" or "no separating lines" but keep the scaling off.

### DIFF
--- a/src/plugins/compact-view/compact.less
+++ b/src/plugins/compact-view/compact.less
@@ -12,9 +12,11 @@
       max(var(--bracket-font-size-factor), var(--minimum-font-size))
     );
   }
-  .dcg-main,
+  .dcg-main {
+    font-size: calc(1.1 * (16 / 18) * var(--math-font-size));
+  }
   .dcg-evaluation-container {
-    font-size: var(--math-font-size);
+    font-size: calc((16 / 18) * var(--math-font-size));
   }
 
   &.compact-view-no-separating-lines {


SlimRunner:
Also, my bad for making so many PRs with small changes, but they are all disjoint in both the files they change and the things being changed. However, if it is preferred to have them in a single one I can close the PRs and rebase them all into a single or fewer ones as needed.

jared-hughes:
Keeping them separate is good! There's essentially no overhead to considering them separately.

jared-hughes:
I like this change, but I'm slightly concerned about migration from the old behavior to the new behavior. Someone who settles on a good font size before this PR would suddenly have the main font 10% bigger. (this isn't negligible: I often accidentally leave a 110% browser zoom on, and it unnerves me).

Brainstormed some ideas. If I were you, I'd probably go with B because it's quick, and meets your goal of the default value matching vanilla.

Idea A:
- keep the default as 18
- divide by 1.1 in the evaluation instead of multiplying by 1.1 for main.

This gives an effective 18/1.1 = 16.36 by default, which is 2% bigger than the default 16. But it's nice because the only thing changing is the evaluation.

Idea B:
- keep the default as 18
- multiply by 1.1 * (16 / 18) in the main, and multiply by (16 / 18) for evaluation

This would shrink people's current sizes by 2%, which isn't too bad. The advantage over A is that the default of 18 now corresponds to exact vanilla.

Idea C: migration
- create a new setting `mathFontSize2` whose default is 16.
- if a settings object has a `mathFontSize` without a `mathFontSize2`, then delete the `mathFontSize` and let the `mathFontSize2` be `mathFontSize * 16 / 18`.
- use `mathFontSize2` in all calculations. 

This matches Idea B in that people's current sizes only shrink by 2%, and the default matches the vanilla sizing before _and_ after. One disadvantage is you can get decimals. If the current size is 19, then it would become 16.889, which looks funny in the input box. But rounding to one decimal place would be reasonable.

Idea D: lazy
- write a PR comment saying "nah people can just adjust the config"
- merge

SlimRunner:
Oh I see, I did not consider that. Yeah, B sounds good. I will work on that one.

SlimRunner:
Uh, sorry forgot to revert the default.

SlimRunner:
B is implemented!

jared-hughes:
Nice, and lovely commit comment!

PR #935, opened 2024-10-05
Fix symbol replacement for better evaluation

SlimRunner:
Better evaluation had partially broken because of two replacements. This fixes the issue.

### DIFF
--- a/src/plugins/better-evaluation-view/better-evaluation-view.replacements
+++ b/src/plugins/better-evaluation-view/better-evaluation-view.replacements
@@ -118,12 +118,12 @@ Add a value field to "undefined" labels
 
 *Find* => `undefined`
 ```js
-if ($options = $options || {}, isNaN($val) || !isFinite($val)) return { type: "undefined" }
+if (isNaN($val) || !isFinite($val)) return { type: "undefined" }
 ```
 
 *Replace* `undefined` with
 ```js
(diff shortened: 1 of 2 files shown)


SlimRunner:
Just a quick breakdown of why it broke.

Both replacements happen inside the same function which is referred to as `$numericLabel` within the replacements. The `$options` variable is no longer used in the function because individual variables are destructured at the beginning presumably only the ones used in this function. As a result the `$options` object is no longer used directly.

SlimRunner:
Another replacement broke today (thanks for @Heavenira for helping me notice it).

It is the one that allows DesModder to duplicate the metadata it uses for pinning expressions and enable GLesmos.

src/plugins/better-evaluation-view/better-evaluation-view.replacements
```diff
@@ -118,12 +118,12 @@ Add a value field to "undefined" labels
 
 *Find* => `undefined`
 ```js
-if ($options = $options || {}, isNaN($val) || !isFinite($val)) return { type: "undefined" }
+if (isNaN($val) || !isFinite($val)) return { type: "undefined" }
 ```
 
 *Replace* `undefined` with
 ```js
-if ($options = $options || {}, isNaN($val) || !isFinite($val)) return { type: "decimal", value: isNaN($val) ? "\\mathrm{NaN}" : $val === Infinity ? "\\infty" : $val === -Infinity ? "-\\infty" : "undefined"}
```
jared-hughes, line 126: Weird that the bundler/minifier put `$options = $options || {},` inside the if statement. Now that it's gone, good riddance I guess

jared-hughes (approved):
Looks good

PR #936, opened 2024-10-09
Prevent panic plugins showing twice in Menu

### LINKED ISSUE #933: Plugins can be disabled multiple times when they share a common error
An image would explain it much better

image

The issue is that unless all components which share the same error are force disabled, Desmodder will still show the entire group of panicked errors including those which have already been force disabled. This other image illustrates the case where only one related item is not force disabled

image

SlimRunner:
This fixes #933

As noted in the issue above, force-disabled plugins which came in "groups" (related by dependencies) would essentially be able to be disabled an infinite number of times. The culprit was that the dependencies were not filtered by the list of force-disabled plugins.

I tested tested this by purposefully breaking random replacements with multiple dependencies. It seemed to work fine in all my tests. However, I am not sure yet if there is a cleaner way to achieve the same (without using `window` or `as any`), but it is currently just four lines.

### DIFF
--- a/src/panic/panic.ts
+++ b/src/panic/panic.ts
@@ -1,4 +1,4 @@
-import { Console } from "../globals/window";
+import window, { Console } from "../globals/window";
 import { format } from "#i18n";
 import { PluginID } from "../plugins";
 import { Block } from "../preload/replacementHelpers/parse";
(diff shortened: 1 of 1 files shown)


jared-hughes (approved):
Looks fine. If it works, it works. The current behavior of editing the DOM _during_ the processing is weird. I'd prefer pushing to a list (or adding to a set), then creating the DOM all at once at the end. Then it's easy to de-duplicate by using a Set.

PR #938, opened 2024-10-16
Fix broken call to get graph title

SlimRunner:
The function `getCurrentGraphTitle` broke because a property changed in the `calc` object. This PR fixes the issue.

It is a very minor issue, but I thought of making it a PR before I forget.

Funny enough, I [fixed this same issue in my old user-script](https://github.com/SlimRunner/desmos-scripts-addons/commit/5a9fe674d85ee6abff42aabf5fcdcc83e1237e14) about a month ago, and even though I got that snippet from fireflame, I did not connect the dots until today that it was likely broken here too.

### DIFF
--- a/src/globals/Calc.ts
+++ b/src/globals/Calc.ts
@@ -166,7 +166,7 @@ export interface TimingData {
 }
 
 export interface TopLevelComponents {
-  headerController: {
+  mygraphsController: {
     graphsController: {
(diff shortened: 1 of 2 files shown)


jared-hughes (approved):
Thanks! For reference, one way this problem appears is in producing a default filename or Video Creator exports.

PR #941, opened 2024-10-24
Paste Image Plugin

64qq:
Added a plugin for pasting images from clipboard (#847).
Feature: When the user paste images
- with an expression selected, they are inserted after it
- with no expression or focusable element selected, they are inserted at the end of the expressions list
- no-op otherwise

I appreciate ronwnor helping me with learning event handling in desmos.


ronwnor:
I've played around with the plugin for a bit and it feels sooooo good! this is amazing :heart: 

a few small problems though:
- pasting an image when the last expression is not empty, it ends up selecting that expression (now one-before-last expression) after pasting
- trying to paste plain text when no expression is selected should do nothing, but when the plugin is enabled, it focuses the last expression

ronwnor:
i think the 2nd bug could be fixed by moving the first if statement inside the second one
the 1st one could maybe be fixed by dispatching another `set-focus-location` event after pasting..? that feels clunky though, there might be better ways idk

src/plugins/paste-image/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  afterEnable() {
+    document.addEventListener("paste", this.pasteHandler);
+  }
+
+  afterDisable() {
+    document.removeEventListener("paste", this.pasteHandler);
+  }
+
+  _pasteHandler(e: ClipboardEvent) {
+    if (!this.calc.selectedExpressionId) {
```
ronwnor, line 18: place this `if` statement inside `if (clipboardFiles?.length){ ... }` to prevent any interaction when there are no images to paste

src/plugins/paste-image/index.ts
```diff
@@ -0,0 +1,42 @@
...
+      // avoid images being inserted at the top of the expressions list
+      // when there is no selected expression
+      if (document.activeElement !== document.body) return;
+      const lastExpression: Desmos.ExpressionState = this.calc
+        .getExpressions()
+        .splice(-1)[0];
+      this.cc.dispatch({
```
ronwnor, line 25: replacing this dispatch event with `dispatch({ type: "new-expression-at-end" })` might be better?
removes the need for `lastExpression` and prevents a selection bug

need to add it as a valid event in `./Calc.ts` though

src/globals/Calc.ts
ronwnor, line 1: add `| "new-expression-at-end"` on line 36 (as a type for `VanillaDispatchedEvent`): see other comments

ronwnor (changes requested):
tried using the review feature (thanks slimrunner) to put comments closer to the relevant code.

src/plugins/paste-image/index.ts
64qq: I didn't realize that. Thanks for your advice!

src/plugins/paste-image/index.ts
64qq: That way is simpler, but that sometimes makes strange the way of inserting images.
- It inserts new expression even when the last expression is empty
- When inserting images, the insertion location is determined for each, rather than following the first insertion location inside `_toplevelNewItemAtSelection`.  In the method of using `"new-expression-at-end"`, the second and subsequent images are inserted at the top due to the specification that inserting an image with an empty expression selected replaces the expression with the image and then deselect.

example

I noticed during this process the bug that images aren't inserted when `showIDs` is enabled and `getExpression()` returns `[]`.

64qq: Sorry I guess I was mistaken. The latter also happens when using `lastExpression`... Anyway, I found a solution to the problem of the insertion of multiple images, but there are still a few problems. I'd like some more time.

jared-hughes:
This looks great! Beautiful behavior. Why is it still in draft? Just checking you have no pending concerns. I saw something about it not working with Show IDs, but I can't reproduce that. Seems to be fine.

I'm impressed that pasting multiple images works well and easily. Looks like the images are pasted in reverse order though.
image

src/plugins/paste-image/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  _pasteHandler(e: ClipboardEvent) {
+    const clipboardFiles = e.clipboardData?.files;
+    if (clipboardFiles?.length) {
+      e.preventDefault();
+      if (!this.calc.selectedExpressionId) {
+        // avoid images being inserted at the top of the expressions list
+        // when there is no selected expression
+        if (document.activeElement !== document.body) return;
+        const lastExpression = this.calc.getExpressions().splice(-1)[0];
```
jared-hughes, line 25:
```suggestion
        const lastExpression = this.calc.getExpressions().at(-1);
```

src/plugins/paste-image/index.ts
```diff
@@ -0,0 +1,41 @@
...
+        lastExpression &&
```
jared-hughes, line 26: Better to make the control flow more obvious.
```suggestion
        if (lastExpression)
```
 `&&` would be good on an assignment RHS like `const id = lastExpression && lastExpression.id` (though that's typically equivalent to `const id = lastExpression?.id`. Would only need a `&&` if it's a construction like `const id = lastExpression && normalizeID(lastExpression.id)`).

64qq:
I have a local implementation of a function to align the order in which images are pasted. It's a trivial thing, but because I use `Object.groupBy` in some of that code, I need to use typescript 5.7. Committing with `--no-verify` is fine, but it's not that urgent, so I've been working on upgrading ESLint and typescript recently. I've got the opportunity to have them reviewed and will now incorporate suggested changes and push them.

64qq:
I'm aware of problems with files randomly changing order, at least in Windows. This comes from ClipboardEvent, so it might be a good idea to sort it on our side.
Should we give the option of leaving it as it is or sorting it? Not sure if it's worth it.

jared-hughes:
> files randomly changing order

What, like rearranging to a random permutation? That doesn't sound like easy to fix.

I thought the problem was that the images were consistently in reverse order, at which point it's reasonable to just reverse the list. But if it's more, it might not be worth fixing. Most people won't copy-paste several images at once (I was surprised that it was possible in the first place).

64qq:
I got it. I was using [Files](https://github.com/files-community/files) for testing, but with the Windows' default explorer.exe it works fine. It might be a Files issue (at least, not a Windows one)
Even without that, in the first place I hadn't thought about the practical matter. It was redundant to provide such an option. However, there may be users like me, so would it be better to sort them implicitly in a common sense order (ascending order with name as key)?

jared-hughes:
Actually, the current behavior of pasting here matches what happens when you drag-drop several images in at the same time. So it's a Desmos issue that it gets reversed.

The random ordering sounds like a Files bug in this case (and we needn't put a workaround just for it). The ordering should really match the order that the files appear in the file manager, which may not necessarily be alphabetical.

src/plugins/paste-image/index.ts
```diff
@@ -0,0 +1,105 @@
...
+      }
+    };
+    const callbackId = this.cc.dispatcher.register(handleImageUpload);
+  }
+
+  static fileListOf(...files: File[]): FileList {
+    const dT = new DataTransfer();
+    files.forEach((file) => dT.items.add(file));
+    return dT.files;
+  }
```
jared-hughes, line 104: It should fine to just pass a `File[]` to Desmos instead of needing a `FileList`.

64qq:
> Actually, the current behavior of pasting here matches what happens when you drag-drop several images in at the same time. So it's a Desmos issue that it gets reversed.
> 
> The random ordering sounds like a Files bug in this case (and we needn't put a workaround just for it). The ordering should really match the order that the files appear in the file manager, which may not necessarily be alphabetical.

Indeed, it may not be the DesModder's job to go that far. I chose not to implement it.
As for lint & test, I'm about to create a new PR including upgrading typescript and typescript-eslint.

src/plugins/paste-image/index.ts
64qq: I didn't notice. Thank you!

jared-hughes (approved):
Awesome, thanks for contributing and updating dependencies along the way!

PR #950, opened 2024-11-28
Fix broken replacements for code golf UI

SlimRunner:
Maintenance PR to fix the UI for code golf. The replacements to show the stats in each expressions broke.

### DIFF
--- a/src/plugins/code-golf/code-golf.replacements
+++ b/src/plugins/code-golf/code-golf.replacements
@@ -12,6 +12,7 @@ $DCGView.createElement("div", {
     class: ()=>({
         "dcg-do-not-blur": !0,
         "dcg-expressionitem": !0,
+        "dcg-highlighted-expressionitem": this.controller.isExpressionHighlighted(this.id),
         "dcg-mathitem": !0,
         __rest0__
         }),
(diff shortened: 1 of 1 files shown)


SlimRunner (commented):
Whenever this is reviewed, I just want to point out that the following warning appears sometimes when a graph with folders just loaded:
```
VM3531:48 Error: Trying to update view that is not mounted. Ignoring update. [FolderCostPanel]
View Hierarchy:
<SwitchUnion>
  <b0>
    <If>
      <FolderCostPanel>
    at vb (eval at load (script.ts:81:7), <anonymous>:48:44)
    at FolderCostPanel.update (eval at load (script.ts:81:7), <anonymous>:49:587)
    at FolderCostPanel.recalculate (index.tsx:173:10)
    at index.tsx:184:12
    at ga.__trace__ (eval at load (script.ts:81:7), <anonymous>:99:1863)
```

I do not know if it is something that should happen, or if it happened even before this fix. I tested as thoroughly as I could and code golf seems to work as intended with the fix. If this is something that needs addressing let me know.

jared-hughes:
I can't imagine that problem could be caused by this PR. So I'll merge.

PR #952, opened 2024-11-30
Deps: Bump typescript to v5.7

64qq:
This PR bumps these packages to the latest version:

* typescript v5.4.3 -> v5.7.2
* typescript-eslint v8.15.0 -> v8.16.0

Includes the following changes:

* Remove no longer needed type assertions
  Path-sensitive type inferences has been added for `Array.prototype.every` and `Array.prototype.filter` ([typescript@5.5](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#inferred-type-predicates))
* Add non-null assertion to `cachedGolfStatsPool.keys().next().value` 
  `BuiltinIteratorReturn` is now changed from `any` to `undefined` in projects with `strict: true` ([typescript@5.6](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#strict-builtin-iterator-checks-(and---strictbuiltiniteratorreturn)))
* Enable `allowImportingTsExtensions` and `emitDeclarationOnly` ([typescript@5.7](https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/#path-rewriting-for-relative-paths))
* Explicitly include or reference files outside the project
  TypeScript language service no longer search outside the project for files that are not explicitly included ([typescript@5.7](https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/#faster-project-ownership-checks-in-editors-for-composite-projects))

### DIFF
--- a/text-mode-core/tsconfig.json
+++ b/text-mode-core/tsconfig.json
@@ -2,13 +2,12 @@
   "extends": "../tsconfig-base",
   "compilerOptions": {},
   "references": [
+    // https://github.com/microsoft/TypeScript/issues/33685
+    // { "path": "../src" },
     { "path": "../localization" },
     { "path": "../metadata" },
-    { "path": "../parsing" }
+    { "path": "../parsing" },
+    { "path": "../graph-state" }
   ],
-  "include": [
-    "**/*",
-    "../src/tests/run_calc_for_tests.ts",
-    "../loaders/utils.mjs"
-  ]
+  "include": ["**/*", "../src", "../loaders/utils.mjs"]
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,8 +56,8 @@
         "puppeteer": "^22.6.0",
         "ts-jest": "^29.1.2",
         "tsutils": "^3.21.0",
-        "typescript": "^5.4.3",
-        "typescript-eslint": "^8.15.0"
+        "typescript": "^5.7.2",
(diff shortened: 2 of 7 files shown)


text-mode-core/tsconfig.json
```diff
@@ -2,13 +2,12 @@
...
-    { "path": "../parsing" }
+    { "path": "../parsing" },
+    { "path": "../graph-state" }
   ],
-  "include": [
-    "**/*",
-    "../src/tests/run_calc_for_tests.ts",
-    "../loaders/utils.mjs"
-  ]
+  "include": ["**/*", "../src", "../loaders/utils.mjs"]
```
jared-hughes, line 12: Checking my understanding: could you confirm if this comment is right?

The current project structure is unfortunate here --
- `src` references `text-mode-core` of course to do text-mode stuff.
- `text-mode-core` references `src` just for `src/tests/run_calc_for_tests`.
- Entirely reasonably, cyclic references are not allowed (https://github.com/microsoft/TypeScript/issues/33685 as you note), so the easiest way to make this work is to say that `src/tests/run_calc_for_tests` is part of both projects.

Maybe it would be better to split off `src/tests` into its own project. But that's out of the scope of this PR.

jared-hughes:
I'm just going to merge to unblock https://github.com/DesModder/DesModder/pull/941. The only thing this PR #952 could break is lint or tests, and it's not breaking them.

PR #959, opened 2024-12-01
Add a scroll event listener to `view.scrollDOM` that closes the item settings menu

### LINKED ISSUE #958: Text Mode: Scrolling the editor does not close the item settings menu
In normal mode, the currently open item settings menu closes when scrolling the expression list. The same behavior is preferred in text mode.

64qq:
Closes #958
The event listener will be removed when unmounting the editor for compatibility. Usually GC works so I'm not sure if it is always necessary.

### DIFF
--- a/src/plugins/text-mode/index.ts
+++ b/src/plugins/text-mode/index.ts
@@ -40,12 +40,21 @@ export default class TextMode extends PluginController {
     this.cc.updateViews();
   }
 
+  abortController = new AbortController();
+
   /**
    * mountEditor: called from module overrides when the DCGView node mounts
    */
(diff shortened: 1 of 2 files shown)


src/plugins/text-mode/index.ts
```diff
@@ -40,12 +40,19 @@ export default class TextMode extends PluginController {
     this.cc.updateViews();
   }
 
+  abortScrollListenerController = new AbortController();
```
jared-hughes, line 43: Ooh I haven't seen AbortController before. Nicer way to abort the listener than the old API

jared-hughes (approved):
Clean fix.

PR #977, opened 2025-01-12
Fix Github workflow failures

Naitronbot:
Hoping that this magically fixes the ci integration test issues.


jared-hughes (approved):
Thanks! Pinning a specific version is good since we'll be able to update it on our terms instead of whenever github decides to update it.

PR #980, opened 2025-01-18
Add showPerformanceMeter to Calculator Settings

Naitronbot:
Not sure if this is the right place for this, may be better to put it in Performance Display.


jared-hughes:
Seems a reasonable place for me. Performance Display feels more like just that specific popover.

PR #984, opened 2025-02-25
Fix broken replacements for Better Evaluation View

64qq:
- `this.props.val()` is not working anymore, so I replaced them with `this.getEvaluationRHS().val` or `itemModel.formula.typed_constant_value.value`. For lists, the latter is not affected by truncation and always return all the elements.
- Added a replacement for `emptyList`
- Wrapped `StaticMathQuillView` with `div.dcg-evaluation-view__wrapped-value` to apply evaluation view styles

### DIFF
--- a/src/plugins/better-evaluation-view/better-evaluation-view.int.test.ts
+++ b/src/plugins/better-evaluation-view/better-evaluation-view.int.test.ts
@@ -2,7 +2,7 @@ import { Driver, clean, testWithPage } from "#tests";
 
 async function expectEval(driver: Driver, latexExpected: string) {
   const latexFound = await driver.$eval(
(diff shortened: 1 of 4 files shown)


64qq:
I have to make it work with expressions without assignments

jared-hughes:
Looks good! Thanks for the fixes. I'll merge with the build CI failing and fix the upload-artifact version in a separate PR.

Regarding

> I have to make it work with expressions without assignments

I just updated the tests so that's not strictly necessary. The choice of `[1,2,3]` not showing output (but `[1,2,3]+0` yes) is intentional by Desmos because you can just see the list right there, no need to show the evaluation output. Same as why just writing `3` ina  line doesn't show output.

PR #993, opened 2025-03-08
Upgrade prettier to v3

64qq:
Upgraded prettier from v2.8.8 to v3.5.3 and applied some code style fixes. 

Lint failure is caused by `trailingComma` now defaulting to `all`, which was `es5` until v2. I set it to `es5` for now, as running prettier as is would result in 120+ more file differences. If we should adopt this change, just fix them and then merge. Otherwise create a prettier config file with `trailingComma` set to `es5`. As you can see, trailing commas are added in type parameters despite opting out of this setting (b361339151259093fecc3ffbff7b638289e0ee18), but this is unavoidable (refer to https://github.com/prettier/prettier/issues/15142).

As for .prettierignore, I did nothing particular. Prettier now takes .gitignore into account in addition to .prettierignore, but it is fine as is. Of course, I'll leave these preferences to maintainers.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,7 +52,7 @@
         "less": "^4.2.0",
         "minimist-lite": "^2.2.1",
         "node-fetch": "^3.3.2",
-        "prettier": "^2.8.8",
+        "prettier": "^3.5.3",
         "puppeteer": "^24.0.0",
         "ts-jest": "^29.1.2",
         "tsutils": "^3.21.0",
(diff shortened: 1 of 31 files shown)


jared-hughes (approved):
Thanks for the updates! I'll hold off on updating `trailingComma` to the default of `all` just because I don't feel like doing it. Doesn't seem much harmful though.

Note .prettierignore is subtly different from .gitignore, so the .prettierignore can't be removed in its entirety.

PR #994, opened 2025-03-13
Fix broken replacements for override-keystroke

64qq:
Just added one line to fix replacement error for override-keystroke.


jared-hughes (approved):
Nice, thanks.

PR #996, opened 2025-03-18
Fix most broken replacements

SlimRunner:
There are some changes that need review. I did not have time to thoroughly test them all. I left comments in the changes that I noticed, but there might be some others I did not catch.

### DIFF
--- a/src/core-plugins/expr-action-buttons/expr-action-buttons.replacements
+++ b/src/core-plugins/expr-action-buttons/expr-action-buttons.replacements
@@ -12,7 +12,7 @@ Insert before the delete button but after the other buttons.
 
 *Find*
 ```js
-.createElement(
+$['n'](
   'span',
   {
     class: () => ({
@@ -21,10 +21,10 @@ Insert before the delete button but after the other buttons.
     ),
     ____
   },
-  $DCGView.createElement(____),
-  $DCGView.createElement(____),
-  $DCGView.createElement(____),
-  $DCGView.createElement(____),
+  $DCGShared['n'](____),
+  $DCGShared['n'](____),
+  $DCGShared['n'](____),
+  $DCGShared['n'](____),
   __delete_button__
 )
 ```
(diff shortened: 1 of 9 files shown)


src/plugins/better-evaluation-view/better-evaluation-view.replacements
```diff
@@ -19,17 +19,17 @@ undefined: ()=> $Dcgview.createElement(
...
+undefined: () => $DcgShared.Components.IfElse(
     () => DSM.betterEvaluationView,
     {
-      true: () => $Dcgview.createElement(
+      true: () => $DcgShared['n'](
         Desmos.Private.Fragile.StaticMathquillView,
         {
           latex: () => this.getNumberLabelString?.() || "undefined",
-          config: $Dcgview.const({})
+          config: {}
```
SlimRunner, line 29: I this the correct way to write this? I did it based on the fact that most constants are now bare.

src/core-plugins/expr-action-buttons/expr-action-buttons.replacements
64qq, line 1: `['n']` depends on certain build of source code, so it might not work with the next build

SlimRunner:
the `Tweak placement of existing settings menu in geometry` in `pillbox-menus.replacements` was left broken. It could not find a way to fix it right now. I will be able to continue working on this branch on Wednesday.

That replacement places the DesModder buttons inline when in Geometry. It does not work. Probably it needs more careful editing.

src/core-plugins/expr-action-buttons/expr-action-buttons.replacements
```diff
@@ -12,7 +12,7 @@ Insert before the delete button but after the other buttons.
 
 *Find*
 ```js
-.createElement(
+$['n'](
```
64qq, line 15: `['n']` depends on certain build of source code, so it might not work with the next build

jared-hughes:
Closed in favor of #999.

Thanks for the kickstart here. This also helped the Desmos team notice that, due to a bundler misconfiguration, the createElement stuff was sent to the worker instead of just the main page. Which is why it was in `__dcg_shared_module_source__['n']` rather than minified to a single variable like `x`.

PR #997, opened 2025-03-18
Use Node.js v22 in GitHub Actions and update DEVELOPMENT.md

64qq:
Set `node-version` to 22.x in workflows/lint-and-test.yml and updated DEVELOPMENT.md accordingly. (Node.js v18 is going to be EOL at the end of April.)
This PR also allows several ECMAScript features (~~like import attributes~~This was my misunderstanding. It is backported to v18) to be available.


jared-hughes (approved):
:+1:

PR #998, opened 2025-03-18
Move scripts/eslint-rules as a separated project and rewrite rules in TypeScript

64qq:
Flat Config made it easier to create and manage custom rules. This PR creates a separated project `@desmodder/eslint-rules` (using npm workspaces) with the following changes:
* Rewrote rules in TypeScript.
  Some conditional branches were removed because they were determined to be unnecessary by type checking. Also, the `meta.docs.category` property has been removed as it is deprecated.
* Dropped tsutils from dependencies and replaced with its successor [ts-api-utils](https://github.com/JoshuaKGoldberg/ts-api-utils) instead.

notes:
* `ESLintUtils.RuleCreator` requires docs url. I set it to the url to the source code of the rule for now, but it also makes sense to write simple docs for each rule in README.md or something.
* `compilerOptions.module` is now set to `ESNext`, but this won't affect because module bundling is handled by esbuild.

#993 and #997 were created because import attributes were needed while preparing this PR.


eslint-rules/package.json
```diff
@@ -0,0 +1,37 @@
...
+    ".": {
+      "import": {
+        "types": "./dist-ts/index.d.ts",
+        "default": "./dist/index.js"
+      }
+    }
+  },
+  "scripts": {
+    "build": "tsc --build && node esbuild.mjs",
+    "postinstall": "npm run build"
```
jared-hughes, line 18: Nice, so this'll automatically build on `npm install`. I was worried there would be an extra dev step.

eslint-rules/index.ts
```diff
@@ -0,0 +1,29 @@
...
+} satisfies FlatConfig.Plugin;
+
+const allRules = {
+  "@desmodder/eslint-rules/no-expect-promise": "error",
+  "@desmodder/eslint-rules/no-external-imports": "error",
+  "@desmodder/eslint-rules/no-format-in-ts": "error",
+  "@desmodder/eslint-rules/no-reach-past-exports": "error",
+  "@desmodder/eslint-rules/no-timeouts-in-intellisense": "error",
+} satisfies Record<
+  `${typeof PLUGIN_NAME}/${keyof typeof rules}`,
```
jared-hughes, line 20: Love this verification that the keys all start with the plugin name.

eslint-rules/rules/no-external-imports.ts
```diff
@@ -13,13 +14,14 @@ module.exports = {
     },
     schema: [],
   },
+  defaultOptions: [],
   create: function (context) {
     if (!context.filename.includes("text-mode-core")) return {};
     return {
       ImportDeclaration: function (node) {
-        if (node.source.type === "Literal" && node.importKind !== "type") {
+        if (node.importKind !== "type") {
```
jared-hughes, line 22: Surprised the `source` is typed only as `Literal` and not `Literal | Identifier`, ref [estree ImportDeclaration spec](https://github.com/estree/estree/blob/35aa0852c62696cf578f65caf129313c789c8057/experimental/module-declarations.md?plain=1#L32).

Removal looks fine I guess.

jared-hughes (approved):
Looks good. Review comments are just idle chit-chat.

I assume the integration test failure is either a flaky test or something changing on Desmos's side. Will come back to this after checking out the other pending PRs.

tsconfig-base.json
```diff
@@ -12,7 +12,7 @@
     "noUnusedParameters": true,
 
     "target": "es2020",
-    "module": "es6",
+    "module": "ESNext",
```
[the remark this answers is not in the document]
64qq: Yes, exactly.

PR #1001, opened 2025-03-20
Fix broken replacements for `Show list elements`

64qq:
- Added replacements for `list-count`


jared-hughes (approved):
Thanks, good fix. It shows `[1,2,3]` when a literal list `[1,2,3]` is entered.

PR #1002, opened 2025-03-20
Fix broken replacements for GLesmos

64qq:
Fixed `_graphFunc` replacements

### DIFF
--- a/src/plugins/GLesmos/glesmos.replacements
+++ b/src/plugins/GLesmos/glesmos.replacements
@@ -112,7 +112,7 @@ $createElement(
   'div',
   {
     class: 'dcg-options-menu-content',
-    children: __children__
+    children: [__children__]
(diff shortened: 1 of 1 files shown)


64qq:
oops

64qq:
It was found that the line confirm replacements is not working properly ([ref](https://discord.com/channels/655972529030037504/655972529030037507/1353223063499706408)). I have tried digging into it but have not yet found a solution

jared-hughes:
I found and pushed a fix to that.

jared-hughes:
Ah it didn't properly fix, it just fixed one symptom. I'm working on a proper fix though -- surprisingly two replacements were replacing the same code and it got confusing when one failed but not the other.

jared-hughes (approved):

PR #1003, opened 2025-03-21
Fix broken replacements for Right Click Tray

64qq:
Replaced `$DCGView.addCustomAttribute` with a single identifier `$addCustomAttribute`


jared-hughes (approved):
Nice and simple. Thanks.

PR #1010, opened 2025-03-26
Fix broken replacements for Better Evaluation View

64qq:
Fixed replacements: `list-count`, `list` and `emptyList`


jared-hughes (approved):
Thanks! This decreases the number of replacement errors, and code changes look good, so I assume it's a step in the right direction.

PR #1011, opened 2025-03-29
Fix/broken-replacements-2025-03

SlimRunner:
I fixed and tested most warnings with broken replacements. There is still an issue with the pillbox. It succeeds in placing the menu below, but the buttons are not placing themselves into the toolbar. That may be a separate issue, I think?


SlimRunner:
I tried to fix the pillbox issue, but I cannot seem to figure it out.

SlimRunner:
Force pushed to remove fix which is already covered by #1010.

jared-hughes (approved):
Thanks for the fixes! Not sure what you mean by the pillbox problems. It's looking fine to me.

PR #1012, opened 2025-04-02
Better-nav-hide-scroll-bar

SlimRunner:
Add setting to show or hide scrollbar for long expressions when better navigation is enabled.

As [discussed here](https://discord.com/channels/1095245998747557978/1095245999473176709/1356759569283551404), it may be convenient to hide it in devices with trackpads and/or touch screen.


jared-hughes (approved):
Looks good! I've inverted the meaning of the checkbox so it actually corresponds to showing the scrollbar.

Thanks for the translation updates too.

This is a somewhat new CSS feature (https://caniuse.com/mdn-css_properties_scrollbar-width), but that's OK since it's opt-in and can fall back gracefully to just not hiding the scrollbar.

PR #1016, opened 2025-04-19
Fix issue with list of points in Better Evaluation View

### LINKED ISSUE #1015: List of points displays as list of complex
Better evaluation view is incorrectly displaying points as complex. https://www.desmos.com/calculator/idt0mddimt

Image

They're represented the same way as a runtime value; need to inspect the valueType.

64qq:
Fix #1015
- Extracted `NonfolderItemModelBase["formula"]` types to interface
  - Made property `expression_type` and `is_inequality` present only in `ExpressionFormula`
  - Made property `is_inequality` optional since it exists only in inequalities as far as I could find
- Added typings for `typed_constant_value`
- Return `itemModel.formula.typed_constant_value` instead of `itemModel.formula.typed_constant_value.value` in `listEvaluation` callbacks so that we can access to the `valueType` property
- Added tests for `EmptyList`, `ListOfComplex`, `ListOfPoint`, and `ListOfPoint3D`
- Made `expectEval` and `expectEvalPlain` into methods of `Driver` to deal with multiple expressions in one test case

to-do:
- Add more detailed typings for `Private`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: npm install and build
         run: |
           npm ci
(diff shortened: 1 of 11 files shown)


jared-hughes (approved):
Looks great, thanks for the fixes and the tests! :sparkling_heart:

64qq:
Oh, I forgot to remove the old replacements

PR #1019, opened 2025-04-19
Add support for exporting frames as WebP

64qq:
With webp selected, ffmpeg will now be run with `-vcodec libwebp -lossless 1 -loop 0` and generate an animated WebP file.

Tweaked PillboxMenu styles to prevent the "Select export type" control from overflowing.


jared-hughes:
Considering splitting the picker into two lines like the points picker
image

and using that extra space to add icons as follows:
- gif, apng, webp get `.dcg-new-image`: dcg-new-image since they're 'image' formats (designed for short animations, suitable for the  tag)
- mp4, webm get `.dcg-film`: dcg-film since they're 'video' formats (designed for videos, suitable for the <video> tag)
- zip gets a folder icon.

But on further reflection I think the distinction will be lost on most people. Not going to try it.

jared-hughes (approved):
Nice

PR #1021, opened 2025-04-23
Fix startup errors

64qq:
Fixed GLemos replacements and PillboxContainer style. 
I'm not too sure but I think it is fine without additional stylings for PillboxContainer? `getPillboxBackgroundColor` seems to have been inlined into the `--dcg-custom-background-color-shaded` variable.


jared-hughes:
Closing in favor of #1022 which cherry-picks the first commit but fixes the style issue differently.

PR #1024, opened 2025-04-23
Automatically suggest using the Prettier extension for VSCode

cootshk:
- Set the default formatter to Prettier in the workspace settings
- Suggest installing Prettier if it isn't installed already
- Update the docs to match


jared-hughes (approved):
Nice

PR #1029, opened 2025-05-06
Dark Theme

gradylink:
I was working on https://github.com/catppuccin/userstyles/pull/1731 when I realized this could also be a DesModder plugin, so here it is! It is still missing support for various DesModder plugins, but I am slowly adding support. I'm going to leave this as a draft at least until https://github.com/catppuccin/userstyles/pull/1731 is merged to give them a chance to find any issues with the userstyle. Currently, the colors are pretty bad (I'm terrible at choosing colors) and I really should have used variables, so I will add those later today or tomorrow. If anyone wants to help pick colors, please do because I really suck at it.

## Plugin Support

- [x] Text Mode BETA
- [x] Compact View
- [x] Intellisense
- [ ] Syntax Highlighting
- [x] The DesModder Menu
- [x] Video Creator
- [x] Performance Display
- [ ] Set Primary Color (IDK how this one would work)
- [ ] Show Tips
- [x] Multiline Expressions

### DIFF
--- a/localization/en.ftl
+++ b/localization/en.ftl
@@ -132,7 +132,7 @@ builtin-settings-opt-zoomButtons-name = Show Zoom Buttons
 builtin-settings-opt-zoomButtons-desc = {""}
 builtin-settings-opt-keypad-name = Show keypad
 builtin-settings-opt-keypad-desc = {""}
-builtin-settings-opt-showPerformanceMeter-name = Show performance meter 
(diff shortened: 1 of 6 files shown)


gradylink:
I think I might swap out the current system of just one singular color scheme with letting the user select from a few (catppuccin variants, gruvbox, dracula, etc.)

PR #1030, opened 2025-05-07
Don't insert pasted images into special folders

64qq:
This PR fixes the behavior of Paste Image so that pasted images won't be inserted into special (and secret) folders such as the previous metadata and the geometry folder.

### DIFF
--- a/metadata/manage.ts
+++ b/metadata/manage.ts
@@ -1,5 +1,8 @@
 import Metadata, { Expression } from "./interface";
 
+export const ID_METADATA = "dsm-metadata";
+export const ID_METADATA_FOLDER = "dsm-metadata-folder";
+
 export function getBlankMetadata(): Metadata {
   return {
     version: 2,
--- a/src/plugins/paste-image/index.ts
+++ b/src/plugins/paste-image/index.ts
@@ -1,7 +1,8 @@
 import { PluginController } from "../PluginController";
 import { format } from "#i18n";
-import type { DispatchedEvent } from "#globals";
+import type { DispatchedEvent, ItemModel } from "#globals";
 import type { AtLeastOne } from "#utils/utils.ts";
+import { ID_METADATA_FOLDER } from "#metadata/manage.ts";
 
 export default class PasteImage extends PluginController {
   static id = "paste-image" as const;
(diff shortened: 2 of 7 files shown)


src/plugins/paste-image/index.ts
```diff
@@ -48,11 +49,25 @@ export default class PasteImage extends PluginController {
...
-      if (!this.calc.isAnyExpressionSelected) {
-        if (document.activeElement !== document.body) return;
-        // Avoid images being inserted at the top of the expressions list when there is no selected expression
-        const lastExprId = this.calc.getExpressions().at(-1)?.id;
-        if (lastExprId) this.setFocusLocation(lastExprId);
+      const geoFolderId = this.cc.getGeometryFolder()?.id;
+      const folderIdDenyList = [ID_METADATA_FOLDER].concat(geoFolderId ?? []);
+      const isInOrIsSpecialFolder = (model: ItemModel) =>
+        folderIdDenyList.includes(model.id) ||
+        (model.folderId && folderIdDenyList.includes(model.folderId));
```
jared-hughes, line 56: The right logic here seems to be instead
```suggestion
const isInOrIsSpecialFolder = (model: ItemModel) => model.isHiddenFromUI
```
This way, you can't paste into _any_ secret folders normally, and you can never paste into the geometry folder. Also, if authorFeatures are enabled, you can paste into all secret folders.

Note `model.isHiddenFromUI` is mostly `model.secret && !cc.shouldShowAuthorFeatures()` but with a few more quirks. In particular collapsed folders' children are marked as `model.isHiddenFromUI = true`. I think that's reasonable enough. If a collapsed folder is selected, we'd paste after the entire folder.

metadata/manage.ts
```diff
@@ -1,5 +1,8 @@
 import Metadata, { Expression } from "./interface";
 
+export const ID_METADATA = "dsm-metadata";
+export const ID_METADATA_FOLDER = "dsm-metadata-folder";
+
```
jared-hughes, line 5: Love the refactoring here so everyone can access the constants, + using this to get rid of some magic strings.

tsconfig-base.json
```diff
@@ -13,7 +13,7 @@
 
     "target": "es2020",
     "module": "ESNext",
-    "lib": ["ES2024.Object"],
+    "lib": ["ES2024.Object", "ES2023.Array"],
```
jared-hughes, line 16: I assume this is for findLast. Seems good enough usage for me, https://caniuse.com/mdn-javascript_builtins_array_findlast.

jared-hughes (changes requested):

src/plugins/paste-image/index.ts
64qq: Wow, this is a really elegant solution! Thank you for your suggestion.

jared-hughes (approved):
Nice fix!

PR #1043, opened 2025-05-13
Support additional value types in BEV list-count

64qq:
Fixes crashes when typing lists of polygons, segments, vectors, and such.

- Added detailed types for `Desmos.Private `(a2863e9209459a5b94c1eef61418ed9072bac447) and refactored the label formatting process in a type-safe way and use iterators when mapping formatter functions (e75339c91ddf6eccfd454b9166b26e816406cc30). 
  This change would be useful in the event of a future need to expand the list of values, since iterator mappings are evaluated lazily. These two commits were cherry-picked from branch no-explicit-any (which is still a work in progress).
- Removed the old replacements (7caf5e2d348baab72137806237a0612c144ed0c3).
- Added extra types for `ValueType` and `ValueTypeMap`. Not all types could be deduced, so I need to look into this more.
- Support for displaying list values of additional types. Still wondering what method would be best as for `AngleMarker`, `DirectedAngleMarker` and `Transformation`.

### DIFF
--- a/src/plugins/better-evaluation-view/components/ListEvaluation.tsx
+++ b/src/plugins/better-evaluation-view/components/ListEvaluation.tsx
@@ -1,54 +1,85 @@
 import { jsx } from "#DCGView";
 import { StaticMathQuillView } from "#components";
-import { Private, TypedConstantValue, ValueType } from "#globals";
+import {
+  LabelOptionsBase,
+  Private,
+  TypedConstantValue,
+  ValueType,
+  ValueTypeMap,
+} from "#globals";
 
-export function ListEvaluation(
-  val: () => TypedConstantValue<
-    | ValueType.EmptyList
-    | ValueType.ListOfNumber
-    | ValueType.ListOfComplex
-    | ValueType.ListOfPoint
-    | ValueType.ListOfPoint3D
-  >
+const labelOptions = {
+  smallCutoff: 0.00001,
+  bigCutoff: 1000000,
+  digits: 5,
+  displayAsFraction: false,
+} satisfies LabelOptionsBase;
+
+const { complexNumberLabel, pointLabel, point3dLabel, truncatedLatexLabel } =
+  Private.Mathtools.Label;
+
+type ListValueType =
+  | ValueType.EmptyList
+  | ValueType.ListOfNumber
+  | ValueType.ListOfComplex
+  | ValueType.ListOfPoint
+  | ValueType.ListOfPoint3D;
+
+type TypedConstantIteratorValue<T extends ListValueType> = T extends T
+  ? {
+      valueType: T;
+      iterator: ArrayIterator<ValueTypeMap[T][number]>;
+    }
+  : never;
+
+function formatLabels<T extends ListValueType>(
+  typedConstantValue: TypedConstantValue<T>
 ) {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+  const { valueType, iterator } = {
+    valueType: typedConstantValue.valueType,
+    iterator: typedConstantValue.value.values(),
+  } as TypedConstantIteratorValue<T>;
+  switch (valueType) {
+    case ValueType.ListOfComplex:
+      return iterator.map((label) => complexNumberLabel(label, labelOptions));
+    case ValueType.ListOfPoint:
+      return iterator.map((label) => pointLabel(label, labelOptions));
+    case ValueType.ListOfPoint3D:
+      return iterator.map((label) => point3dLabel(label, labelOptions));
+    default:
+      return iterator.map((label) => truncatedLatexLabel(label, labelOptions));
+  }
+}
+
+// https://github.com/tc39/proposal-iterator-helpers/issues/296
+// The result of the iterator helpers is an object with prototype %IteratorHelperPrototype%, which is a closable iterator.
+// So take(n) implicitly calls the 'return' method and closes the iterator. This is unexpected when intending to access to the rest of the iterator.
+function* reusableTake<T>(iterable: Iterator<T>, limit: number) {
+  let i = 0;
+  while (i++ < limit) {
+    const result = iterable.next();
+    if (result.done) break;
+    yield result.value;
+  }
+}
+
+export function ListEvaluation(val: () => TypedConstantValue<ListValueType>) {
   return (
     <div class="dcg-evaluation-view__wrapped-value">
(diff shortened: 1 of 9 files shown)


64qq:
It seems I need to add support for inference-related value types

src/plugins/better-evaluation-view/components/ListEvaluation.tsx
```diff
@@ -1,58 +1,185 @@
...
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+  const { valueType, iterator } = {
+    valueType: typedConstantValue.valueType,
+    iterator: typedConstantValue.value.values(),
+  } as TypedConstantIteratorValue<T>;
+  switch (valueType) {
+    case ValueType.ListOfComplex:
+      return iterator.map(complexNumberLabel);
+    case ValueType.ListOfAny:
+      return iterator.map(String);
```
jared-hughes, line 79: I suspect this is unreachable; if Desmos starts producing ListOfAny, this could end up giving `[Object object]` type stuff. Doubt there's a better idea though.

src/plugins/better-evaluation-view/components/ListEvaluation.tsx
```diff
@@ -1,58 +1,185 @@
...
+      return iterator.map((bool) => `\\mathrm{${bool}}`);
+    case ValueType.ListOfPoint:
+      return iterator.map(pointLabel);
+    case ValueType.ListOfPoint3D:
+      return iterator.map(point3dLabel);
+    case ValueType.EmptyList:
+      return iterator;
+    case ValueType.ListOfPolygon:
+      return iterator.map(
+        (points) => `polygon\\left(${points.map(pointLabel).join(",")}\\right)`
```
jared-hughes, line 92: Nice, this works for empty polygons

src/plugins/better-evaluation-view/components/ListEvaluation.tsx
```diff
@@ -1,58 +1,185 @@
...
 
 export function ListEvaluation(
-  val: () => TypedConstantValue<
-    | ValueType.EmptyList
-    | ValueType.ListOfNumber
-    | ValueType.ListOfComplex
-    | ValueType.ListOfPoint
-    | ValueType.ListOfPoint3D
-  >
+  typedConstantValue: TypedConstantValue<NormalListValueType>
```
jared-hughes, line 163: I'm surprised this works. Normally, passing around a value (rather than a callback to get the value) doesn't work out in DCGView because the DOM doesn't auto-update. But if it works, probably fine to leave it be.

src/plugins/better-evaluation-view/index.ts
```diff
@@ -1,24 +1,64 @@
...
+      case ValueType.RGBColor:
+        return settings.colors
+          ? (swatch) => ColorEvaluation(value, swatch)
+          : undefined;
+      case ValueType.ListOfColor:
+        return settings.colors && settings.lists && settings.colorLists
+          ? (swatch) => ColorEvaluation(value, swatch)
+          : undefined;
+      default:
+        return settings.lists ? () => ListEvaluation(value) : undefined;
```
jared-hughes, line 48: This makes me suspicious because it implies that the cases excluded above guarantee that `value` has type `NormalListValueType`. But if Desmos ever adds a new list type, then this assumption can break. In this case, I know it won't _crash_ because `ListEvaluation` only relies on the type in `formatLabels` which has a safe `default` case, but it'll still look weird.

But I'd rather it be the other way around: have a bunch of cases (`case ListOfNumber:`, `case ListOfComplex:` etc) that limit the type of `value` to a `NormalListValueType`, then let the `default` case be the `return undefined` that `ListOfChiSquareIndependence` etc should reach. Then it would work even if you never thought about `ListOfChiSquareIndependence` in the first place.

src/preload/replaceElement.ts
```diff
@@ -31,10 +31,14 @@ export function insertElement(creator: () => undefined | (() => any)) {
...
+  key: () => unknown = () => !!replacer()
+) {
   const { DCGView } = (Desmos as any).Private.Fragile;
-  return DCGView.Components.IfElse(() => !!replacer(), {
-    true: () => replacer()!(old()),
-    false: () => old(),
-  });
+  return createElementWrapped(DCGView.Components.Switch, {
+    key,
+    children: () => (replacer() ?? ((_) => _))(old()),
```
jared-hughes, line 42:
```suggestion
    children: () => (replacer() ?? ((x) => x))(old()),
```

tsconfig-base.json
```diff
@@ -13,7 +13,7 @@
 
     "target": "es2020",
     "module": "ESNext",
-    "lib": ["ES2024.Object", "ES2023.Array"],
+    "lib": ["ES2024.Object", "ES2023.Array", "ESNext.Iterator"],
```
jared-hughes, line 16: How well-supported is Iterator? I don't know what to search on https://caniuse.com/. 

Looks like this PR works in my Chrome and Firefox browsers, so it's probably fine (DesModder doesn't support Safari)

jared-hughes (changes requested):
Looks mostly fine, but I'm nervous about being confident in any exhaustive list of types. For example, if you remove all references to `ListOfChiSquareIndependence`, you get a mess: https://www.desmos.com/calculator/egaktnmid0. Any future type Desmos adds could suffer the same fate.

So it's better in general to make the cases we support explicitly opt-in, rather than excluding every case that doesn't work.

tsconfig-base.json
64qq: caniuse: <https://caniuse.com/mdn-javascript_builtins_iterator_map>
I think it's okay to consider the Baseline as an indicator of whether a feature is acceptable to use in production, though Iterator Helpers are pretty new (Baseline 2025).

src/plugins/better-evaluation-view/components/ListEvaluation.tsx
jared-hughes: Aha. Does that key help performance?

64qq: That is for subscribing to the `typed_constant_value` changes rather than improving the performance.
As `BetterEvaluationView#evaluation` (formerly `#listEvaluation` and `#colorEvaluation`) conditionally returns `undefined` or a replacer component depending on the value type now, we need a way to rerun the function on value update.

PR #1054, opened 2025-06-25
Added "Quake Pro" Feature

Heavenira:
This feature introduces an ability to increase the Field of View of the 3D calculator beyond the regular limit.

The name comes from FOV setting in Minecraft, where if you maximize the viewport, its named as "Quake Pro". Can rename it to anything you like, i found this catchy.

The method is to multiply anything related to `perspectiveDistortion` when it comes to rendering. This will retain vanilla behavior, albeit with a higher magnification limit.

(Thanks to @SlimRunner for teaching me how to do replacements!!!)

feature screenshot

### DIFF
--- a/src/plugins/quake-pro/index.ts
+++ b/src/plugins/quake-pro/index.ts
@@ -1,0 +1,26 @@
+import { PluginController } from "../PluginController";
+
+export default class quakePro extends PluginController {
+  static id = "quake-pro" as const;
+  static enabledByDefault = true;
+
+  oldName = "";
+
+  afterEnable() {
+    const headerElement = getHeaderElement();
+    if (headerElement === null) return;
+    const text = headerElement.innerText;
+    if (text !== undefined) this.oldName = text;
+    headerElement.innerText = "DesModder ♥";
+  }
+
+  afterDisable() {
+    const headerElement = getHeaderElement();
+    if (headerElement === null) return;
+    headerElement.innerText = this.oldName;
+  }
+}
+
+function getHeaderElement(): HTMLElement | null {
+  return document.querySelector(".header-account-name");
+}
--- a/src/plugins/quake-pro/index.ts
+++ b/src/plugins/quake-pro/index.ts
@@ -6,6 +6,7 @@ export default class quakePro extends PluginController<Config> {
   static enabledByDefault = true;
   static config = configList;
 
+  magnification = 1;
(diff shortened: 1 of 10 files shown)


src/plugins/quake-pro/quake-pro.replacements
```diff
@@ -0,0 +1,43 @@
...
+  let $e = this.graphSettings.config.perspectiveDistortion;
+  return __return__
+}
+```
+
+*Replace* `from` with
+```js
+{
+  let $e = this.graphSettings.config.perspectiveDistortion;
+  return __return__ * (DSM.quakePro?.magnification ?? 1);
```
jared-hughes, line 22: The function currently looks like
```
        getPerspectiveDistortion() {
            let e = this.graphSettings.config.perspectiveDistortion;
            return isNaN(e) ? 1 : e < 1e-4 ? 0 : clamp(e, 0, xk)
        }
```
The replacement is token-based, so it will only multiply the clamp:
```
        getPerspectiveDistortion() {
            let e = this.graphSettings.config.perspectiveDistortion;
            return isNaN(e) ? 1 : e < 1e-4 ? 0 : clamp(e, 0, xk) * (DSM.quakePro?.magnification ?? 1);
        }
```

Is this what you expect? Maybe the 1 should be multiplied by the magnification, which would be putting parentheses around the return.
```suggestion
  return (__return__) * (DSM.quakePro?.magnification ?? 1);
```

src/plugins/quake-pro/index.ts
```diff
@@ -0,0 +1,31 @@
+import { PluginController } from "../PluginController";
+import { Config, configList } from "./config";
+
+export default class quakePro extends PluginController<Config> {
+  static id = "quake-pro" as const;
+  static enabledByDefault = true;
```
jared-hughes, line 6:
```suggestion
  static enabledByDefault = false;
```

jared-hughes (changes requested):

Heavenira:
@jared-That was unintentional by me!! I hadn't considered that the ternary was swallowing the multiplication.

This should fix an issue I noticed upon page loading. There was a single frame with an incorrect FOV value. I suspect because `e` was `NaN` it was ignoring the multiplication. I hope now this is prevented.

jared-hughes (approved):
Awesome! Merging...

PR #1057, opened 2025-06-28
Add empty braces after `\infty` to avoid unexpected parsing of the command

64qq:
image

Currently, you can't use `i * ∞` to represent `i * ∞` because it evaluates to `NaN + NaNi` (I assume that's because desmos can't distinguish `i * ∞` from `(0 + i) * ∞`), but `√-∞` can be used instead. However, the evaluation view shows an empty box because the latex `\inftyi` returned by the label function is invalid.

This PR fixes the behavior of `numericLabel` to correctly show the evaluation for complex numbers with an imaginary part of +/- infinity.


jared-hughes (approved):
Nice fix. A single space would work too, but braces are equivalent.

PR #1061, opened 2025-07-04
Migrate find-replace view management to DCGView

### LINKED ISSUE #1051: Ticker appears under pinned expression after using Ctrl+F
1. Pin an expression
2. Set up the ticker
3. Ctrl+F
4. Exit Ctrl+F

Ticker is now below the pinned expression

Image

64qq:
This will fix #1051 and the problem where nesting of the search bar container gets deeper and deeper if you repeat opening and closing the expression search.


jared-hughes:
Hah, I wasn't aware of the nesting issue. It is funny.

image

jared-hughes (approved):
Nice fix. Old code must've been one of the first times we replaced DOM stuff. The new code is much nicer to read -- it's more obviously correct

PR #1062, opened 2025-07-05
fix styling for usage with custom colors / background colors

ronwnor:
modifies hardcoded color values so they don't look off with custom colors


src/core-plugins/pillbox-menus/components/Menu.less
```diff
@@ -80,7 +80,7 @@
 
 .dsm-plugin-info-body {
   padding: 5px 0;
-  color: #444;
+  color: var(--dcg-custom-background-color-shaded, #444);
```
64qq, line 83:
```suggestion
  color: var(--dcg-custom-text-color-shaded, #444);
```
Did you mean this?

PR #1063, opened 2025-07-06
Fix the remaining coloring issues

64qq:
This PR adds coloring fixes missing in #1062

Background colors with alpha channels like
https://github.com/DesModder/DesModder/blob/1190ffc830eb5f0ef68898d4ec5c11c78af60a2c/src/plugins/syntax-highlighting/index.less#L1-L4
are those that will be the original colors when mixed with the default background color.


src/plugins/compact-view/compact.less
```diff
@@ -1,6 +1,11 @@
+// https://stackoverflow.com/a/28592238
+.layered-background(@a, @b: var(--dcg-item-background-color)) {
+  background: linear-gradient(@a), linear-gradient(@b);
```
jared-hughes, line 3: Cool trick!

jared-hughes (approved):

PR #1065, opened 2025-07-17
Update Japanese localizations

64qq:
- Added Japanese translations for Quake Pro and BEV floats.
- Renamed '改行つき数式' (Multiline) to '複数行モード', which is more concise & less clunky.


jared-hughes (approved):
Nice!

PR #1067, opened 2025-07-17
Set the maximum width of the current action to avoid overflowing

64qq:
| Before | After |
| --- | --- |
| before | after |


jared-hughes (approved):
Looks good. This scrolls horizontally even if "Better Navigation" is disabled, which is correct behavior since you can't navigate with the mathquill cursor.

PR #1068, opened 2025-07-18
Place the golf panel under the evaluation dropdown

64qq:
| Before | After |
| --- | --- |
| before | after |


jared-hughes (approved):
Nice fix

PR #1086, opened 2025-10-09
Added Zed support, refactored Wolfram2Desmos

Definitely-Not-A-Dolphin:


jared-hughes:
Closing in favor of https://github.com/DesModder/DesModder/pull/1091, which takes the same wolfram2desmos refactors and adds tests. I went with a unit test (.unit.test.ts) rather than an integration test (.int.test.ts) because the tests didn't need to interact with the DOM, since most of wolfram 2 desmos is inside a single unit-testable function.

The tests caught one bug, a typo fixed in https://github.com/DesModder/DesModder/pull/1091/commits/ed6f5bc18d2f7caced79f41306a4070a92ed1f68.

If you want to continue with adding Zed support, I'm not strongly opposed. Just make sure it's a PR that only contains the Zed support so it can be considered in isolation. Also Feny's comments are good, in particular look at the association for `.replacements` files like `src/plugins/code-golf/code-golf.replacements` -- they should be syntax highlighted as Markdown files.

PR #1088, opened 2025-10-18
Graph library plugin

Lenny-the-burger:
Simple graph library plugin

image

Doesn't store the actual graph, only the url. When you click on the one you want it copies the link to clipboard and then you can paste it into an empty folder title and Desmos handles the importing.

Uses local storage because I dont know what else to use.

Long titles are truncated. If you have a graph that starts with `@private` it removes the @ and makes it bold (in the list view). This is for better integration with the intellisence plugin (folder gets marked as private once imported).

Full disclosure this was mostly slopped together by ai, but it was either that or I don't do it at all and I'm tired of this not being a thing. Let me know if something is broken, everything seems to work on my end.


Lenny-the-burger:
Ill have to go and resetup some env stuff probably, I ran npm run lint and it told me they there were 300 files with issues which probably isn't right

src/plugins/graph-library/index.ts
```diff
@@ -0,0 +1,211 @@
...
+  getCurrentGraphUrl(): string {
+    return window.location.href.split("?")[0].split("#")[0]; // Remove query params and hash
+  }
+
+  getCalculatorTypeFromUrl(url: string): "graphing" | "geometry" | "3d" | "unknown" {
+    if (url.includes("/calculator/")) return "graphing";
+    if (url.includes("/geometry/")) return "geometry";
+    if (url.includes("/3d/")) return "3d";
+    return "unknown";
+  }
```
Naitronbot: Could pull this from the controller instead of the url

`this.cc.getProduct()`.

src/plugins/graph-library/index.ts
```diff
@@ -0,0 +1,211 @@
...
+        break;
+      case "dsm-graph-library-open":
+        if (action.url) {
+          window.open(action.url, "_blank");
+        }
+        break;
+      default:
+        action satisfies Exclude<DispatchedEvent, AllActions["graph-library"]>;
+    }
+    return undefined;
```
Naitronbot: Don't think this is needed, functions return `undefined` by default

src/plugins/graph-library/index.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  library: GraphLibraryEntry[] = [];
+  copiedEntryId: string | null = null;
+
+  afterEnable() {
+    this.loadLibraryFromStorage();
+    this.dsm.pillboxMenus?.addPillboxButton({
+      id: "dsm-graph-library-menu",
+      tooltip: "graph-library-name",
+      iconClass: "dsm-icon-bookmark",
```
Naitronbot: Thoughts on using the `book` or `books` icon? [Would have to import them from Icomoon](https://github.com/DesModder/DesModder/blob/main/src/fonts/README.md).

src/plugins/graph-library/components/LibraryPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+								</div>
+								<div class="dsm-graph-library-entry-actions">
+									<If predicate={() => this.gl.copiedEntryId === entry().id}>
+										{() => (
+											<i class="dcg-icon-check dsm-graph-library-copied-icon" />
+										)}
+									</If>
+									<If predicate={() => this.gl.copiedEntryId !== entry().id}>
+										{() => (
+											<i
```
Naitronbot: We might want to wrap these in `<button>`. Also we should add tooltips ([example](https://github.com/DesModder/DesModder/blob/main/src/plugins/performance-info/PerformanceView.tsx#L18)).

src/plugins/graph-library/index.ts
```diff
@@ -0,0 +1,211 @@
...
+    this.library.splice(index, 1);
+    this.saveLibraryToStorage();
+
+    // Update the view to reflect the deletion
+    this.dsm.pillboxMenus?.updateMenuView();
+  }
+
+  saveLibraryToStorage() {
+    try {
+      localStorage.setItem(
```
Naitronbot: Should we use extension storage instead of localStorage?

src/plugins/graph-library/index.ts
```diff
@@ -0,0 +1,211 @@
...
+    this.dsm.pillboxMenus?.updateMenuView();
+  }
+
+  async importGraph(entryId: string) {
+    const entry = this.library.find((e) => e.id === entryId);
+    if (!entry) return;
+
+    try {
+      // Copy URL to clipboard
+      await navigator.clipboard.writeText(entry.url);
```
Naitronbot: Thoughts on, instead of copying to the user's clipboard, just directly simulating a paste onto the expression list?

```ts
const data = new DataTransfer();
data.setData('text/plain', entry.url);

const pasteEvent = new ClipboardEvent('paste', {
  bubbles: true,
  clipboardData: data
});

document.querySelector(".dcg-expressionlist").dispatchEvent(pasteEvent)
```

My snippet is a bit of a jank way to do it, but something along those lines.

PR #1089, opened 2025-10-21
Rename `_showToast` to `showToast`

64qq:
`Calc.controller._showToast` was renamed to `showToast`.


jared-hughes (approved):
Nice fix.

Looks like it's also possible to dispatch an action to show a toast, like 
```js
Calc.controller.dispatch({
  type: "toast/show",
  toast: { message: "message" },
});
```
Both of these methods are officially private, so I suspect they'd have equal stability going forward. No strong reason to choose one over the other, other than the dispatch approach forces a view update and can't be called inside a dispatch.

PR #1090, opened 2025-10-21
Replace DesModderFragile components with built-in ones

64qq:
Replaced components derived from DesModderFragile (`ImageIconView`, `ExpressionIconView`, `FooterView`) with built-in Fragile components to fix broken style circles.


jared-hughes (approved):
Thanks for the fix!

PR #1101, opened 2025-10-28
Clear pending timeouts when Intellisense is disabled

64qq:
Currently, some intellisense timeout handlers are called after Calc is destroyed on tool switch, causing errors. This PR adds cleanup for pending timeouts to `afterDisable` to avoid these errors.

### DIFF
--- a/src/plugins/intellisense/index.tsx
+++ b/src/plugins/intellisense/index.tsx
@@ -772,5 +772,8 @@ export default class Intellisense extends PluginController<{
     }
 
     if (this.dispatcher) this.cc.dispatcher.unregister(this.dispatcher);
+
+    pendingIntellisenseTimeouts.forEach(clearTimeout);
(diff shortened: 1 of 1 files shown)


src/plugins/intellisense/index.tsx
```diff
@@ -114,7 +114,7 @@ export default class Intellisense extends PluginController<{
 
   async waitForCurrentIntellisenseTimeoutsToFinish() {
     await new Promise<void>((resolve) => {
-      const currentTimeouts = Array.from(pendingIntellisenseTimeouts.entries());
+      const currentTimeouts = [...pendingIntellisenseTimeouts];
```
64qq: I doubt this function has ever worked properly, since `currentTimeouts` is a list of `[number, number]` but the function compares its elements against `number`.

PR #1143, opened 2026-03-31
Dispatch tick after running showToast (and more)

64qq:
- Add utility functions
  - `tick`: shorthand for `cc.dispatch({ type: "tick" })`
  - `showToast`: utility for displaying a toast and updating the view right after that (so it will be displayed without user action).
    It also accepts toast messages to be localized (formatted), removing the need for using `format` and adding `eslint-disable-next-line` directives.
- Add a new property `util` to `PluginController`. All its members are basically the same as utilities in depUtils.ts that have a `calc` parameter, but they are partially applied with argument `this.calc`.

### DIFF
--- a/localization/i18n-core.ts
+++ b/localization/i18n-core.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @desmodder/eslint-rules/no-format-in-ts */
 import enFTL from "./en.ftl";
 import esFTL from "./es.ftl";
 import frFTL from "./fr.ftl";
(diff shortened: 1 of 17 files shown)


localization/i18n-core.ts
```diff
@@ -13,6 +14,18 @@ export const locales = new Map<string, FluentBundle>();
...
+  key: string;
+  args?: Record<string, FluentVariable> | null;
+  missingReplacement?: string;
+}
+
+export const fromFormattable = ({
+  key,
+  args,
+  missingReplacement,
+}: Formattable) => format(key, args, missingReplacement);
```
64qq, line 27: I'm wondering if we should add this to `no-format-in-ts`

jared-hughes: Added in [e6ca372](https://github.com/DesModder/DesModder/pull/1143/commits/e6ca372793531d64709049a509d0cefd9122807b)

PR #1144, opened 2026-04-06
Add more 3D settings

ninjaguardian:
Supports `lockRotation` and `disableLighting` in text mode.
Unit tests were also added for 3D settings.


jared-hughes (approved):
Nice, this works well, and I appreciate the test!

PR #1151, opened 2026-05-16
Replace DWindow interface with global declaration merging

Naitronbot:


src/globals/window.ts
```diff
@@ -19,20 +19,22 @@ import { GenericSettings, PluginID } from "../plugins";
...
-    pluginsForceDisabled: Set<PluginID>;
-    pluginsEnabled: Record<PluginID, boolean | undefined>;
-    pluginSettings: Record<PluginID, GenericSettings | undefined>;
-  };
-  Desmos: Desmos;
-  shellController: {
-    graphsController: {
-      getCurrentGraphTitle: () => string;
+declare global {
+  interface Window {
```
Naitronbot: This makes it so you still need to do `window.DSM`, as opposed to being able to just directly access `DSM` as a global. We could declare these directly as globals if we wanted.

jared-hughes: I don't see an advantage for directly accessing `DSM` as a global. Bit of conciseness but no value-add.

jared-hughes (approved):

PR #1158, opened 2026-05-23
BEV list length display option

ronwnor:
Replaces better evaluation view's `list` toggle with a segmented option like video creator's export tab. This allows switching between the original list evaluation view, the newer native view, and a minimal view that only shows list lengths.  

preview:
image

I'm not sure about the option names (new -> native?); it would be nicer to have control over this in the localization files I think (which will also enable translating them to other languages). 
It would also be cool to have tooltips for each segment, but I imagine that'd be more complicated.

Speaking of, the setting title/tooltip need to be updated (have been so for a long time). I can push that in this pr if someone with a better english can provide me with a good description.


ronwnor:
I'm not sure if the way I'm updating the view when switching between options (4th commit) is correct, or if there's a better way to do it

ronwnor:
what did I break

SlimRunner:
@ronwnor I think the test breaking may be expected. I'm on mobile so I can't check right now, but check the integration test for BEV. It is failing because the expected output of some previews is not the same anymore.

To check this locally run this

```
npm run test:integration
```

Is this change something you expected?

Also, I recommend running the general test before every commit (`npm run test`) because the commit hook does not run the integration tests.

ronwnor:
oh you're right, the tests had
```js
async function enableBevLists(driver: Driver) {
  await driver.setPluginSetting("better-evaluation-view", "lists", true);
}
```
which was toggling the `lists` boolean. it broke because I changed it from boolean to a string[]. I think I just need to change `true` to `"old"`.

tysm!

src/core-plugins/pillbox-menus/components/Menu.tsx
```diff
@@ -233,6 +237,49 @@ function indentation(level: number, inner: any) {
...
 }
 
+function segmentedOption(
+  pm: PillboxMenus,
+  item: ConfigItem,
+  plugin: SpecificPlugin,
+  settings: GenericSettings
+) {
+  const segmentedItem = item as ConfigSegmentedOptions;
+  const value = () => (settings[item.key] as string) ?? "";
```
jared-hughes, line 247: Typically, local values are red flags when it comes to DCGView, but it seems okay in this case, since none of these values will change. (The `{plugin.config.map((item: ConfigItem) => (` determines `item`, and the rest are references to objects that wouldn't change.)

src/plugins/better-evaluation-view/index.ts
```diff
@@ -95,7 +109,12 @@ export default class BetterEvaluationView extends PluginController<Config> {
     const evaluation = this.evaluation(
       this.getTypedConstantValue.bind(container)
     );
-    return !!evaluation && this.getTypedConstantValue.call(container);
+    return (
+      !!evaluation && [
+        this.settings.lists,
+        this.getTypedConstantValue.call(container),
+      ]
```
jared-hughes, line 116: > I'm not sure if the way I'm updating the view when switching between options (4th commit) is correct, or if there's a better way to do it

Your approach here seems reasonable. Extending the cache key is sensible.

src/plugins/index.ts
```diff
@@ -68,11 +68,26 @@ export interface ConfigItemColorList extends ConfigItemGeneric {
...
+interface SegmentedOption {
+  name: string;
+  /** Localization key, see en.ftl. */
+  i18nKey: string;
+}
+
+export interface ConfigSegmentedOptions extends ConfigItemGeneric {
+  type: "segmented-options";
+  // TODO: possibly figure out a way to make the list const so that the
+  // default is type safe?
```
jared-hughes, line 80: No clue here

jared-hughes (commented):
I pushed two commits adding localization, one commit making it so the old true/false system gets migrated to old/new, and one commit fixing a bug where `rgb(1,2,3)` would be treated as "3 element list".

I'm not loving the names "old" and "new" (though "length" is perfect). The "new" could be called "default" instead (it's what Desmos uses by default), but no clue what the "old" would be called.

Maybe it could be "new"="separate" and "old"="together" ... probably needs some brainstorming.

jared-hughes (approved):
Thanks for the PR!

Leaving the TODO on type safety in there seems fine

PR #1166, opened 2026-06-09
Fix scrollbar overflow in better-evaluation-view

Naitronbot:
Changes to the expression result output broke the scrollbar overflow, so output was getting clipped when "Show list elements" was on.

The HTML/CSS of the expression list output has changed a lot since some of this CSS was first written, from what I can tell a lot of it no longer applies.

image

image

### DIFF
--- a/src/plugins/better-evaluation-view/better-evaluation-view.less
+++ b/src/plugins/better-evaluation-view/better-evaluation-view.less
@@ -1,32 +1,13 @@
 .dcg-container.dsm-better-evaluation-view {
-  // Color Swatch
-  .dcg-color-evaluation .dcg-color-evaluation {
-    vertical-align: middle;
-  }
-
-  .dcg-color-evaluation .dcg-static-mathquill-view {
-    margin-right: 5px;
-  }
-
   .dsm-color-container {
     display: flex;
-    flex-direction: row;
+    gap: 5px;
+    max-width: 100%;
     align-items: center;
   }
 
   // Scrollbar for output
-  .dcg-evaluation-container {
-    max-width: 100%;
+  .dcg-evaluation-view__wrapped-value {
     overflow-x: auto;
-    overflow-y: hidden;
-    // -3px margin is removed from .dcg-evaluation due to clipping. Move it here.
-    margin-top: -3px;
-    > .dcg-evaluation.dcg-do-blur {
-      margin-top: 0;
-      max-width: 100%;
-      .dcg-inline-math-input-view {
-        font-size: 100%;
-      }
-    }
   }
 }


src/plugins/better-evaluation-view/better-evaluation-view.less
```diff
@@ -1,32 +1,13 @@
 .dcg-container.dsm-better-evaluation-view {
-  // Color Swatch
-  .dcg-color-evaluation .dcg-color-evaluation {
-    vertical-align: middle;
-  }
```
Naitronbot, line 5: This doesn't appear to be needed, since `.dsm-color-container` is a flex with `align-items: center`

src/plugins/better-evaluation-view/better-evaluation-view.less
```diff
@@ -1,32 +1,13 @@
...
-
-  .dcg-color-evaluation .dcg-static-mathquill-view {
-    margin-right: 5px;
-  }
```
Naitronbot, line 9: I assumed this originally was to put padding between the list container and the color swatch, but whatever it originally applied to, it doesn't seem to match anything now.

I added `gap: 5px` to the container to replicate this. Even if it's not what this was doing originally, I think it looks nice.

Image

src/plugins/better-evaluation-view/better-evaluation-view.less
```diff
@@ -1,32 +1,13 @@
...
-    overflow-y: hidden;
-    // -3px margin is removed from .dcg-evaluation due to clipping. Move it here.
-    margin-top: -3px;
-    > .dcg-evaluation.dcg-do-blur {
-      margin-top: 0;
-      max-width: 100%;
-      .dcg-inline-math-input-view {
-        font-size: 100%;
-      }
-    }
```
Naitronbot, line 30: `.dcg-do-blur` used to be applied to the color swatch container, which used to have the identifier name in it. This doesn't appear to be needed anymore.

(Screenshot of old Desmos)

Image

src/plugins/better-evaluation-view/better-evaluation-view.less
```diff
@@ -1,32 +1,13 @@
...
+    max-width: 100%;
     align-items: center;
   }
 
   // Scrollbar for output
-  .dcg-evaluation-container {
-    max-width: 100%;
+  .dcg-evaluation-view__wrapped-value {
     overflow-x: auto;
-    overflow-y: hidden;
```
Naitronbot, line 21: This `overflow-y` was added in https://github.com/DesModder/DesModder/commit/f034ef0d1be4cd19510ee09cbffac750773bb683

Can't seem to replicate the original issue

jared-hughes (approved):
Nice, thanks for the fixes and the forensics!

PR #1183, opened 2026-08-20
Add backslash command input to Custom MathQuill Config

physy:
Adds an optional Backslash Commands setting to Custom MathQuill Config.

When enabled, typing `\` opens an editable command input. It accepts complete LaTeX fragments such as `\alpha`, `\sqrt{x}`, and `\frac{a}{b}`, which can be committed with Enter or Tab.

- Shows an inline command preview with caret movement and mouse positioning
- Supports editing with Backspace and cancellation with Escape
- Supports custom command expansions, including `#1`, `#2`, … argument templates
- Keeps the implementation compatible with the new MathQuill
- Adds integration tests


physy:
~~Follow-up: this PR now also adds configurable custom backslash commands.~~
Follow-up: I split configurable custom commands into a [separate review PR](https://github.com/physy/DesModder/pull/1) to make this PR easier to review.

The latest commit also limits built-in command insertion to complete LaTeX fragments. For example, use `\sqrt{x}` or `\frac{a}{b}` rather than a bare `\sqrt` or `\frac`. This removes the `mq.cmd()` fallback, which is not compatible with the new MathQuill implementation.

src/plugins/backslash-commands/backslash-commands.int.test.ts
jared-hughes, line 1: I appreciate the tests!

src/plugins/backslash-commands/index.ts
```diff
@@ -0,0 +1,344 @@
...
+  cursorIndex: number;
+  preview: HTMLElement | undefined;
+  sourceElement: HTMLElement | undefined;
+}
+
+type MathQuillFieldWithLatexWriter = MathQuillField & {
+  write?: (latex: string) => unknown;
+};
+
+/** On JIS keyboards the same character can be reported as `¥`. */
```
jared-hughes: Out of curiosity, on https://www.desmos.com/notebook, does pressing `\` on a JIS keyboard not open the command palette?

jared-hughes:
Is there a reason why this is inside the Custom Mathquill Config plugin, rather than being a plugin of its own?

src/plugins/backslash-commands/backslash-commands.int.test.ts
physy: thanks!

src/plugins/backslash-commands/index.ts
physy: This results in somewhat complicated behavior. By default, macOS JIS keyboards cannot input `\` directly; you need to press `Option` + `¥`. If this is inconvenient, you can change the OS keyboard setting so that pressing the `¥` key inputs `\` instead. A certain number of Japanese users use this setting.

However, for some reason, even with this setting enabled, `event.key` appears to remain unchanged—at least in my environment. As a result, after changing the setting, pressing `¥` in Notebook inputs `\`, but it doesn't open the command palette. Conversely, pressing `Option` + `¥` inputs `¥`, but it opens the palette.

physy:
Fireflame told me on discord that it would probably be better
oh i didn't know you were fireflame

jared-hughes:
Oh! Well it was a decent idea, but considering https://github.com/physy/DesModder/pull/1 will add more config options than just enable/disable, this makes sense to be its own plugin.

physy:
yeah, that makes sense. I'll move it right below the Utility section.

src/plugins/backslash-commands/index.ts
Naitronbot: Might be more robust to listen on the MathQuill input's [`beforeinput` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/beforeinput_event), seems like `InputEvent.data` returns the actual character to be typed, rather than the raw keys.

Not able to test that it works for JIS on mac, but can confirm it properly maps `Alt + Numpad1` on windows to "☺", whereas KeyboardEvent returns raw Numpad1.

(ultimately fireflame's discretion, but if this is not easy to implement, probably not necessary for the first version of this plugin)

physy: Sorry for the delayed response. That makes a lot of sense, so I moved command initiation (and text input as well) to `beforeinput` and removed the special handling for `¥`.

PR #1185, opened 2026-08-26
Update spanish translation

SlimRunner:
Add missing items in ES localization and made a minor grammar revision.


jared-hughes (approved):
Nice, thank!

PR #1202, opened 2026-09-24
Color Input Plugin

ronwnor:
Adds a mathfield to the expression options menu for editing the color directly:

image

I've tested it a little and it should work fine with colors and list of colors, though color maps seem to require some more work then just setting the latex. Setting the latex to a defined color map variable works, but shows an error because `.formula.color_latex_valid` returns `false`.


ronwnor:
should it be enabled by default?

